PR #40, opened 2014-12-18
BZ-1175395 - Add multi Instance completion condition support

mswiderski:


tsurdilo:
Maciej and I talked about this and decided to apply his chances from this PL as a separate commit since it can no longer be automatically merged (due to some changes to MI code in designer). 

This code has been applied to master and 6.2.x branches

PR #47, opened 2015-05-04
initial deployment after GWT upgrade

porcelli:
this includes EAP6.4, Tomcat and WildFly 8.1 distros


jlindop:
Hi Porcelli,
I tried this locally and there's a compilation error in jbpm-designer-standalone:

[ERROR]   The project org.jbpm:jbpm-designer-standalone:6.3.0-SNAPSHOT (/home/jeremy/Documents/gitRepositories/droolsjbpm/jbpm-designer/jbpm-designer-standalone/pom.xml) has 1 error
[ERROR]     'dependencies.dependency.version' for de.benediktmeurer.gwt-slf4j:gwt-slf4j:jar is missing. @ org.jbpm:jbpm-designer-standalone:[unknown-version], /home/jeremy/Documents/gitRepositories/droolsjbpm/jbpm-designer/jbpm-designer-standalone/pom.xml, line 253, column 17

I rebuilt all the droolsjbpm projects, just in case, and still get this error. We should either remove the dependency on de.benediktmeurer.gwt-slf4j:gwt-slf4j:jar or add the version to the bom. I tried removing it and designer builds and runs OK. This may be the right thing to do, as other pom's, e.g. jbpm-form-modeler/jbpm-form-modeler-showcase also look as though they depend on it (looking at mvn dependency:tree), but they don't include it.
Thanks,
Jeremy

porcelli:
Hi @jlindop sorry for that... later today I'll send a PR with such dependency added to droolsjbpm build bootstrap.

PR #52, opened 2015-06-18
BPMSPL-127: Add new dialog for defining Assignments and Input/Output …

jlindop:


jlindop:
Closed because being replaced by new, updated branch.

PR #55, opened 2015-07-16
BPMSPL-127: Add UF dialog for Assignments and Data Inputs/Outputs

jlindop:
PR with bug fixes & enhancements suggested by Tiho.


jlindop:
Close, as replacing with new PR.

PR #57, opened 2015-07-31
BZ1248978: Prevent NPE's in Bpmn2JsonMarshaller.java caused by DataOu…

jlindop:
…tputAssociations referencing renamed or deleted process variables.


psiroky:
@jlindop is this change still needed? If so, would it be possible to add a test case?

jlindop:
@psiroky - No, it's not required - I'll close this PR. Thanks for the reminder!

PR #60, opened 2015-08-14
BZ1244249: Add DateTime property to Timers in Designer

jlindop:


jlindop:
Closed due to revision required.

PR #61, opened 2015-08-20
BZ1249559: Improve undo functionality in IE10 / IE11 by executing com…

jlindop:
…mands in refreshCanvasForIE as a group, rather than separately.

Tiho - will you take a look at this and see what you think. It improves the Undo functionality in IE10/IE11 to make it usable, but it's still not perfect. For each action that's been done in a process, the Undo command can be run twice - once to undo the selection of the selected activity, and once to undo the actual action that was done.


tsurdilo:
Jeremy and I discussed about this and decided to implement a global strategy for this that can help all commands. His implementation here is fine but we will look at the ability to omit all/any commands as needed from the undo stack app wide

PR #64, opened 2015-09-09
refactored duplicated code

jomarko:


yurloc:
Jožo, have you considered creating some tests related to the change?

jomarko:
@yurloc , no I want to fix Bpmn2UnmarshallingTestCase at first. There are the same fails on master and my branch.

tsurdilo:
While I do appreciate the effort and your work to refactor some of the common functionality I don't think it's necessary to use terms as "terrible" or "fails"..there is really no need for that. 

I do agree that we should ad some tests because changes to roundtripping bpmn2 even if its just moving some "terrible" common code into separate methods can possible cause changes in the bpmn2 that need to be caught.

tsurdilo:
I would not use "extract" as the method name. Probably better to use "set" please.

Also the level is somewhat too granular IMO, for example when generating color info it would make more sense to have a single setColorInfo for example method than
extractColorProperty("a"); extractColorProperty("b"), extractFontSize(..."

Also like method names such as "extractNameFlowElement" make little sense to people that read the code, especially if you have for example:

protected void applyEndEventProperties(EndEvent ee, Map<String, String> properties) {
        ee.setId(properties.get("resourceId"));
        extractNameFlowElement(ee, properties);

Simpler name as in "setNodeName" for example would be better here as it would apply to flow elements, artifacts, etc and its confusing to have applyEndEventProeprties then set some "flow element name"

Just as last comment if we do go ahead of choose to refactor as you have it it would be better if we are more flexible to pass for example the proeprty name(s) as well as the namespace into the refactored methods, for example:

private void extractIsSelectable(BaseElement sequenceFlow, Map<String, String> properties) {
        if(properties.get("isselectable") != null && properties.get("isselectable").length() > 0) {
            ExtendedMetaData metadata = ExtendedMetaData.INSTANCE;
            EAttributeImpl extensionAttribute = (EAttributeImpl) metadata.demandFeature(
                    "http://www.jboss.org/drools", "selectable", 

Given the nature of bpmn2 and that we often have to create slightly different changes depending on the node types given the same properties it would be better to pass both the property name "isselectable" as well as the namespace in this case ""http://www.jboss.org/drools" to the method. This would even allow you to further refactor all methods which retrieve data from node extension elements in the future and allow us to create small one-offs when needed to implement certain features.

Just my 2c :)

jomarko:
@tsurdilo I didn't want to offend nobody with words as "terrible" and "fails". If I did, sorry for that. By "fails" I meant testcases, which didn't passed.

I was a little bit investigating. In project is used this library [1] and in these failing tests [2] is assertion, if some TextAnnotation belongs to Artifact's of process. But this is basically not possible, because in used library [1] TextAnnotation class is not child of Artifact class.

I am really not sure, how to solve this.

I will apply your suggestions as soon as possible and I will let you know then.

[1]
from effective pom: <version.org.eclipse.bpmn2>0.7.6-jboss</version.org.eclipse.bpmn2>

[2]
Bpmn2UnmarshallingTestCase.testAssociationBidirectionalUnmarshalling:757 ClassCast
  Bpmn2UnmarshallingTestCase.testAssociationUnidirectionalUnmarshalling:742 ClassCast
  Bpmn2UnmarshallingTestCase.testAssociationUnmarshalling:727 ClassCast org.ecli...
  Bpmn2UnmarshallingTestCase.testTextAnnotationUnmarshalling:373 » NoSuchElement

tsurdilo:
@jomarko great catch! That change was done just recently and was to switch TextAnnotations from Artifacts to FlowElement type so that they also can be included inside a LaneSet. 

I think honestly these test no longer apply and can be removed. 

Thanks again for all your help !

jomarko:
All existing tests have passed. I am going to add testcases for code I have refactored. Then I will mark this PR as completed from my site.

jomarko:
@tsurdilo I think, I am done with this PR. Please review the PR and feel free to give me any feedback.

psiroky:
@tsurdilo any update on this? Can it be merged? I see it is not cleanly mergeable anymore, so that would need to be fixed.

csadilek:
+1, @tsurdilo do you have any objections?

porcelli:
@psiroky @tsurdilo is this still makes sense?

PR #74, opened 2015-11-04
BZ 1277883 

tsurdilo:
@jlindop @csadilek @manstis 

PR for BZ 1277883 - check if global directory is present before trying to create it . Same for the default workitem definition

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/repository/vfs/VFSRepository.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/repository/vfs/VFSRepository.java
@@ -580,140 +580,145 @@ public class VFSRepository implements Repository {
 
         ioService.startBatch(new FileSystem[]{ioService.getFileSystem(URI.create(globalDirPath))});
 
-        ioService.createDirectory(globalDirVFSPath);
+        if ( !ioService.exists( globalDirVFSPath)) {
+            ioService.createDirectory(globalDirVFSPath);
 
-        // dummy form templates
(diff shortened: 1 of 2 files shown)


manstis:
I know the issue, I see the solution.. unfortunately I don't see any tests :(

tsurdilo:
@manstis It is hard to have Designer-specific tests for new project creation within the designer code itself. I can add tests in kie-wb do you think that would be better?

manstis:
This is true, if you're looking at an integration test, but for a unit test you could manually invoke the ```createGlobalDirOnNewProject()``` method and check (1) the global files are only created once and (2) no ```FileAlreadyExistsException``` error is thrown.

csadilek:
Yes, it should be relatively easy to write a unit test for this method. Can you please add the test @tsurdilo?

tsurdilo:
@csadilek @manstis 
test added

tsurdilo:
@csadilek I would like to go through all the tests and fix the assertNotNull where needed since the place/test in question is not related to this commit (is some old test). I think for the second comment the test should fail if that exception is thrown so it's good as it is now I think.

tsurdilo:
Changed the assert as requested and squashed commits into one.

csadilek:
+1, thanks for updating the tests @tsurdilo . @manstis do you want to take a look again, before I merge?

manstis:
+1 to merge.

csadilek:
Thanks!

master: b665841062851844b4f6fc8757555edff58f5049

@tsurdilo, just a reminder if you want this for 6.3.x as well, please send a separate PR.

PR #77, opened 2015-11-09
BZ 1278756 - Isn't possible to move group of elements from container

tsurdilo:
@krisv 

JS update to dragging / dropping logic. added case for cross-container moves while keeping everything else in tact as before.

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/dragdropresize.js
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/dragdropresize.js
@@ -1435,31 +1435,64 @@ ORYX.Core.Command.Move = ORYX.Core.Command.extend({
 		this.plugin		= plugin;
 		this.doLayout = doLayout;
 		this.newParents = [];
+
+		var xcontainermove = true;
+		var xcontainerparent;
+		var xcontainervalueparent;
+		if(moveShapes.length == 1) {
+			xcontainermove = false;
+		} else {
+			for(var k=0; k<moveShapes.length; k++) {
+				var value = moveShapes[k];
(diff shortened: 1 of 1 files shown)


jlindop:
+1

psiroky:
I know these are JavaScript changes, but could we add a test case? There surely must be ways to unit test JavaScript stuff.

tsurdilo:
@psiroky i have added rhino locally for unit testing but we need to first discuss if that is what team wants to do and at this point will take me some time to make sure it all works and create a doc on how to write  the js based unit tests. I could for now create a test case in our Java tests that mimics this if really needed.

porcelli:
@tsurdilo we, as a team, agreed from now on all code has to be tested. unless has a good reason for not

tsurdilo:
@porcelli yes I know however I don't think we agreed on what tool(s) we will use for JS testing within our codebases, have we? I can keep working on adding rhino because I am familiar with if everyone is OK with that. I don't know the estimate atm how long it will take me to get that part working for all parts of the JS code in Designer so I guess until then this PR will stay open...

krisv:
We have to make some decisions on where we want to spend our time on and how.  Building a new test framework for code we are putting into maintenance and where we are shifting strategy in the long run (to be more in line with how the rest of the workbench works) might not be worth the effort.  We discussed this before and decided to make an exception for fixes in the designer JS code, so we can focus on refactoring instead.  I agreed with Christian that I would take care of merging these PRs.

porcelli:
@krisv @tsurdilo so... I wasn't aware of a general agreement that JS wouldn't be tested - I was considering that would be a case (PR) by case. That's fine.

krisv:
Merged http://github.com/droolsjbpm/jbpm-designer/commit/a1577332a

PR #78, opened 2015-11-10
BZ-1278596: - Signal subprocess is never invoked

jesuino:
BZ-1278596: - Signal subprocess is never invoked


jesuino:
Fix for the BZ will include other changes.

tsurdilo:
-1 the lookup against the ID is the correct way to go. the correct approach would be to either port the engine parser back or add backwards compatibility instead of reverting a fix imo

PR #83, opened 2015-11-12
BZ 1279507 - Designer don't enter to read-only mode after opening locked process

tsurdilo:
@krisv master branch PR

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/server/EditorHandler.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/server/EditorHandler.java
@@ -199,11 +199,13 @@ public class EditorHandler extends HttpServlet {
     }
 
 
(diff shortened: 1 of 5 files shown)


csadilek:
+1, @krisv unless you have an objection I will go ahead and merge this as it's quite critical.

On a different note, this outlines very well why we need more test coverage and why we should also think of adding some very basic test harness for JavaScript code (not for the canvas but some for basic logic). This feature worked a few weeks ago, caused a regression on the process instance view which has gone unnoticed. The fix for the process instance view in turn caused this logic to break which again has gone unnoticed for weeks. This is not a complaint, just pointing out why we would benefit from some basic JS testing as well. We will have to maintain this code for a long time.

csadilek:
Thanks! (esp. for the test ;))

master: dc6f84fe493d1c520d180320fdcfb3ccbcaea3a8

PR #85, opened 2015-11-16
BZ1268005: Make sure all characters in constants in datamapper are sa…

jlindop:
…ved correctly.

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
@@ -18,6 +18,7 @@ package org.jbpm.designer.bpmn2.impl;
 import java.io.File;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
 import java.text.DecimalFormat;
 import java.util.*;
 import java.util.Map.Entry;
@@ -7278,7 +7279,12 @@ public class Bpmn2JsonUnmarshaller {
         if (s == null || s.isEmpty()) {
             return s;
         }
-        return s.replaceAll("\\|\\|", "=").replaceAll("##", ",");
+        try {
+            return URLDecoder.decode(s, "UTF-8");
+        }
+        catch (UnsupportedEncodingException e) {
+            return s;
+        }
     }
 
 }
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonMarshaller.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonMarshaller.java
@@ -17,6 +17,8 @@ package org.jbpm.designer.bpmn2.impl;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.util.*;
 import java.util.Map.Entry;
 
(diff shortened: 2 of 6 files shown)


jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
```diff
@@ -7278,7 +7279,12 @@ private String decodeAssociationValue(String s) {
...
             return s;
         }
-        return s.replaceAll("\\|\\|", "=").replaceAll("##", ",");
+        try {
+            return URLDecoder.decode(s, "UTF-8");
+        }
+        catch (UnsupportedEncodingException e) {
+            return s;
+        }
     }
```
csadilek: Is it always a URL that gets encoded here or could it be any String value that you want as UTF-8? Then I would do this instead: s.getBytes( Charset.forName( "UTF-8" ) ). This also has the advantage of not throwing a checked exception....so you can get rid of the catch block.

csadilek:
See my comment above. It seems that the fix is about escaping values in the payload but using a URLEncoder/URLDecoder for it. Wouldn't it be more suitable to use a simple escape utility for json? There should be a couple on the classpath already i.e. StringEscapeUtils.escapeJson or escapeEcmaScript. There's also Errai's Marshallutil.jsonStringEscape....

jlindop:
What's going on in this commit is that on the client side, constants (and the whole Business Process) are represented in JSON, but on the server side, they're represented in BPMN2 XML. In the client side JSON we need to escape some characters which are used as markup in the JSON value which includes the constants, because otherwise they can be wrongly be interpreted as markup. For example, the constant "http://www.test.com/getit?a=1&b=2" might be serialized in a JSON value like this:
"[din]myconstant1=http://www.test.com/getit?a=1&b=2"  (this is a simplified example)
in which the first '=' character is used as markup, but subsequent '=' characters are part of a constant.
BZ1268005 reports bugs related to characters in constants being wrongly interpreted as markup characters.
To avoid the later '=' characters being mistaken for markup, I've URLEncoded constants in the JSON serialization. (Note that it's not just '=' that's used as markup, some other characters are as well).
On the server side, Bpmn2JsonUnmarshaller (which converts the JSON BP to XML) URLDecodes the constants, while Bpmn2JsonMarshaller (XML to JSON) URLEncodes them.
On the client side, constants are URLEncoded in the JSON, so we URLDecode them before showing them to the user, then URLEncode them again when serializing to JSON.

csadilek:
Thanks, @jlindop. My comment still makes sense then. The question is why should we use URL encoding/decoding logic to escape JSON and XML values? Looking at your tests this is used for all kinds of values (not just URLs).

jlindop:
Hi Christian,
The other escape functions you mention, e.g. StringEscapeUtils, wouldn't fix the problem. The markup characters we need to escape in constants are '=' and  ','  - these are 'escaped' by URLEncoding.

I'll run through what's going on...
Client code in designer builds json attribute:value pairs such as this, to represent data assignments:
"assignments": "[din]input1=http://www.things.com&a=4,[din]var2->input2,[din]input3=C:\Windows\file.txt" (this value has 3 DataInputs and their assignments), which uses markup characters =,[]->
When this value is represented in BPMN XML, it's parsed into the various parts and appropriate XML is created. We have some very simple parsing code to split the value into its constituents, typically using String.split. Clearly this isn't the ideal way to parse it, but that's how it's currently done.

Using URLEncoding is a pragmatic solution to fix the bug and I would regard it as low risk. 

As an alternative, we could do something like quoting string constants inside assignment strings, e.g.
"assignments": "[din]input1=\"http://www.things.com&a=4\""
rather than the current
"assignments": "[din]input1=http://www.things.com&a=4"
which would allow constants to contain markup characters, then rewrite the way these strings are parsed on both client and server side.

tbh, I don't mind whether we use URLEncoding or change the json serialization and parsing, as there are pros and cons either way.

csadilek:
Hi @jlindop,

OK, so neither '=' nor ',' actually need escaping in XML or JSON. So, it's our serialization/parsing logic that's incorrect here. If I understand correctly you're saying this is a lower risk change to fix the problem.

So, +1.

jlindop:
Hi @csadilek   Yes, that's right.
Thanks for the discussion - talking about this was useful!

porcelli:
+1 based on discussion

csadilek:
Thanks.

master: 4a949ed0e6d095770b741c2087eefed0e285c6b6

PR #86, opened 2015-11-16
Another round of minor fixes

Hasys:


jbpm-designer-backend/src/main/java/org/jbpm/designer/filter/InjectionConfig.java
```diff
@@ -57,14 +50,21 @@ public synchronized void load(String contextPath) {
...
+                Document doc = parser.parse(is);
+                NodeList rulesConf = doc.getElementsByTagName("rule");
+                rules = new InjectionRules();
+                for (int i = 0; i < rulesConf.getLength(); i++) {
+                    Node ruleNode = rulesConf.item(i);
+                    InjectionRule rule = new InjectionRule(ruleNode, contextPath);
+                    if (rule.isEnabled())
+                        rules.add(rule);
+                }
+                is.close();
```
triceo: You need a ```finally``` block to close the stream correctly.

jbpm-designer-backend/src/main/java/org/jbpm/designer/stencilset/Beautifier.java
```diff
@@ -35,6 +31,7 @@ private String getScriptFromFile(String filename) throws IOException {
         int x = fis.available();
         byte b[] = new byte[x];
         fis.read(b);
+        fis.close();
```
triceo: Again, you'll need a ```finally``` block.

triceo:
@Hasys I see that this PR does a lot of work around closing streams. However, it doesn't close them correctly...

See here: http://stackoverflow.com/questions/2699209/java-io-ugly-try-finally-block

triceo:
You can re-throw the exception if you like. But the important point is that, even when the exception is thrown, you still need to close the streams properly.

tsurdilo:
@Hasys it looks like this PR also does not include the fix for problems in DesignerInjectionFilter introduced in your last PR - is this something that you could please add to this one or you plan to do another?

Hasys:
@triceo @tsurdilo Thank you for your replays. Fixed.

porcelli:
No tests?

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
```diff
@@ -681,37 +683,44 @@ private void deletefile(File f) {
...
-                && (numRead = is.read(bytes, offset, bytes.length - offset)) >= 0) {
-            offset += numRead;
-        }
+            int offset = 0;
+            int numRead = 0;
+            while (offset < bytes.length
+                    && (numRead = is.read(bytes, offset, bytes.length - offset)) >= 0) {
+                offset += numRead;
+            }
+            closeStream(is);
```
triceo: It will be enough to close the streams once, inside the ```finally``` block.

triceo:
Good point, @porcelli - we'll need to add those.

PR #89, opened 2015-11-19
Another round of fixes found by findbug.

Hasys:
Sorry for removing old PR it is my bad. But there is new one with fixed remarks and test coverage.

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/filter/InjectionConfig.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/filter/InjectionConfig.java
@@ -15,12 +15,13 @@
 
 package org.jbpm.designer.filter;
 
+import java.io.InputStream;
 import javax.servlet.ServletContext;
 import javax.xml.parsers.DocumentBuilder;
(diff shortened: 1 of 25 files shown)


jbpm-designer-backend/src/main/java/org/jbpm/designer/filter/InjectionConfig.java
```diff
@@ -37,38 +38,38 @@ public InjectionConfig(ServletContext sc) {
...
+                for (int i = 0; i < rulesConf.getLength(); i++) {
+                    Node ruleNode = rulesConf.item(i);
+                    InjectionRule rule = new InjectionRule(ruleNode, contextPath);
+                    if (rule.isEnabled())
+                        rules.add(rule);
+                }
             }
         } catch (Exception e) {
             // TODO Log
             e.printStackTrace();
```
triceo: Can you fix this TODO?

Hasys: @tsurdilo Hi Tihomir. How do you think is it good enough to use org.slf4j.Logger here? It will be better to use Logger#info or Logger#error?

PR #95, opened 2015-11-23
[BZ#1272196] refactored duplicated code

jomarko:


jomarko:
@tsurdilo please I am not sure if this is "feature or bug"

https://github.com/jomarko/jbpm-designer/blob/master/jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerPopUpPresenter.java#L110-L111

https://github.com/droolsjbpm/jbpm-designer/blob/master/jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerPresenter.java#L570-L572

Is it intention that in the first case there is not put sessionId into the parameters?

jbpm-designer-client/src/test/java/org/jbpm/designer/client/parameters/DesignerEditorParametersPublisherTest.java
```diff
@@ -0,0 +1,132 @@
...
+    public void clearParameters() {
+        parameters = new HashMap<String, String>();
+    }
+
+    @Test
+    public void testProcessSources() {
+        parameters.put(DesignerEditorParametersPublisher.PROCESS_SOURCE_KEY, "value");
+        publisher.publishProcessSources(parameters);
+
+        verify(view, times(1)).publishProcessSourcesInfo("value");
```
jhrcek: Just a small comment on economy of expression

    verify(x, times(1)).something();

 is actually equivalent to the more concise

    verify(x).something()

See http://docs.mockito.googlecode.com/hg/org/mockito/Mockito.html#4
Thought I agree it's a minor thing and also a matter of personal preference, I tend to prefer the more concise variant :-)

tsurdilo:
@jomarko regarding the sessionid - I believe this was added in https://github.com/droolsjbpm/jbpm-designer/commit/bc1341422df74e6d4efe7335737a2740b4697edd#diff-15bbeb3687b34467bd5952773790f708 but not to the popup presenter. I don't think it makes a difference however would not hurt to add

regarding the verify expression - we want to make sure that it does not get called more than once so I think the expression is fine as is

jbpm-designer-client/src/test/java/org/jbpm/designer/client/parameters/DesignerEditorParametersPublisherTest.java
tsurdilo: @jhrcek cosmetics :) I think the way it is now seems fine. From readability of the test where the intent is to make sure the methods gets called and no more than 1 times this reads better.

jomarko:
PR ready for review.

jlindop:
@tsurdilo Tiho - I think this looks OK, but it would be good to get approval from you, since you commented on the PR a while ago.

tsurdilo:
Yes it looks OK but I have a couple of comments maybe we could discuss before.
1. DesignerEditorParametersPublisherViewImpl - Not sure why we need a separate class to publish 3 jsni methods. If we go this way then lets just have a single class that is responsible to publish all jsni stuff we have in the 2 presenters. 

2. In DesignerPresenter there are like 4 separate calls:
            designerEditorParametersPublisher.publishActiveNodes(editorParameters);
            designerEditorParametersPublisher.publishCompletedNodes(editorParameters)             
            designerEditorParametersPublisher.putTimeStampToParameters(editorParameters);
            designerEditorParametersPublisher.putSessionIdToParameters(editorParameters); 

If we are externalizing this part of the logic into its own class then to me would make more sense to have lust
designerEditorParametersPublisher.publish(editorParameters);

and let that class deal with knowing what all needs to go into the editor parameters or not.

just my 2c

porcelli:
@Hasys  @tsurdilo is this still makes sense?

jomarko:
@tsurdilo 
To 1. comment: I think we need this to reduce code repetition.
To 2. comment: You are completely right, I will amend PR soon.

jomarko:
Changes pushed

porcelli:
@tsurdilo mind review?

tsurdilo:
+1

@jomarko thanks

porcelli:
https://github.com/droolsjbpm/jbpm-designer/commit/ca2a1dd4a97770626756baf963e4ad7bfa18b596

PR #98, opened 2015-11-27
fix for frame scrollbars

tsurdilo:
@csadilek PR master only (this should not go into other branches yet)

### DIFF
--- a/jbpm-designer-client/src/test/java/org/jbpm/designer/client/shared/OnResizeTest.java
+++ b/jbpm-designer-client/src/test/java/org/jbpm/designer/client/shared/OnResizeTest.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.designer.client.shared;
+
+import com.google.gwt.user.client.ui.FlowPanel;
+import com.google.gwt.user.client.ui.Widget;
+import com.google.gwtmockito.GwtMock;
+import com.google.gwtmockito.GwtMockitoTestRunner;
+import org.jbpm.designer.client.DesignerPopUpPresenter;
+import org.jbpm.designer.client.DesignerViewImpl;
+import org.jbpm.designer.client.DesignerWidgetPresenter;
+import org.jbpm.designer.client.DesignerWidgetView;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.mockito.Spy;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.when;
+
+@RunWith(GwtMockitoTestRunner.class)
+public class OnResizeTest {
+
+    @Spy
+    private DesignerWidgetView designerWidgetView;
+
+    @Spy
+    private DesignerPopUpPresenter designerPopupPresenter;
+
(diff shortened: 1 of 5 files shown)


tsurdilo:
@jlindop can you please review? Thanks

jbpm-designer-client/src/test/java/org/jbpm/designer/client/shared/OnResizeTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Spy
+    private DesignerWidgetView designerWidgetView;
+
+    @Spy
+    private DesignerPopUpPresenter designerPopupPresenter;
+
+    @Spy
+    private DesignerViewImpl designerViewImpl;
+
```
csadilek: This (DesignerViewImpl) is not being used. I guess you wanted to write a test for it too?

jbpm-designer-client/src/test/java/org/jbpm/designer/client/shared/OnResizeTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @GwtMock
+    DesignerWidgetPresenter designerWidgetPresenter;
+
+    @Test
+    public void testDesignerWidgetViewOnResize() {
+        when(parentWidget.getOffsetWidth()).thenReturn(100);
+        when(parentWidget.getOffsetHeight()).thenReturn(100);
+
+        when(designerWidgetView.getParent()).thenReturn(parentWidget);
+
```
csadilek: These 3 when statements are now duplicated and could be moved to a setup method.

tsurdilo:
@csadilek amended PR with the extra test. Moving the 3 when statements to @Before did not work as tests failed so I left them there.

jlindop:
@tsurdilo 
The Process Dialog looks better, but seems to have a misshapen frame. See image:
processmodeldialog

tsurdilo:
@jlindop thats the popup presenter in the process state view. I wonder if there is actually something wrong with that popup. seems way smaller than before.

tsurdilo:
@jlindop I amended a change to popup presenter and the corresponding test. can you please give it another go and test ? Thanks!

jlindop:
@tsurdilo 
To verify we're looking at the same thing, I bring up the dialog by deploying a project with a BP, then go to:
ProcessManagement->ProcessDefinitions, Select my BP, then in the right-hand-frame, select Options->Process model. (Some other ways of bringing up the process model dialog that used to work don't seem to work right now).
I'm still seeing the frame as before. If the main browser isn't very wide, the dialog looks like it's not all showing.
processmodeldialog2

tsurdilo:
@jlindop yes the issue seems to be with the "Process Model" popup which is not part of Designer. If you revert all changes here the result is the same. I am looking into the console code to see what updates the underlying popup may need and will create a separate bz for that but for this scrollbar fix it seems that the issue you are showing is not related to Designer code.

jlindop:
@tsurdilo Ah, I see. Right now, the designer part of this dialog seems big - using my default size browser, I need to make the browser bigger to show the whole process-diagram and the scrollbars around it.
processmodeldialog3

tsurdilo:
@jlindop yes - the parent container has to (and used to) control the displayed with and users can vertically scroll through the process. This no longer seems the case and that's why I said we should look at that popup widget to see what changed / needs to change instead of focus on the designer fix here that really has nothing to do w/ the problem.

jlindop:
@tsurdilo 
I see - thanks for the update. This looks ok then.
+1

csadilek:
@tsurdilo before I merge, is there a BZ for this fix? If so, can you add it to the commit message. It looks a little terse as it doesn't say what it fixes.

jomarko:
@csadilek I think, there is no BZ or JIRA, because I warned @tsurdilo about this issue just by email, I considered it as the fastest way how to get the issue fixed

csadilek:
@jomarko yes it is the best/fastest way.

Thanks.

master: bacb3b750a1538650434a56ee7257dc0b8210aef

manstis:
But someone, IMO, should have raised a JIRA to record the work. Normally we
try *not* to work from emails/IRC chat.

csadilek:
Yes @manstis is right. We should have created a JIRA and referenced it in the commit message then.

jomarko:
@csadilek @manstis I am sorry for this, In future I will always file the issue to JIRA or BZ. In this case I decided for email because I found the issue in one of the first 6.4.0 SNAPSHOT.

PR #101, opened 2015-12-03
BZ 1277522 - Opening Process Sources twice breaks the modal

tsurdilo:
@jlindop - can you please test? Link to BZ https://bugzilla.redhat.com/show_bug.cgi?id=1277522
To reproduce open process, select view process sources and *without* closing the first source window open another one then click+drag it around -- there should not be another blue box appearing behind it from the first window (which is now being destroyed). 

Yes no test and yes I am almost done with adding test framework for JS.

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/view.js
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/view.js
@@ -1922,7 +1922,11 @@ ORYX.Plugins.View = {
                     var winWidth = Math.min(600, visibleWidth * 2/3);
                     var winHeight = Math.min(550, visibleHeight * 2/3);
 
-                    var win = new Ext.Window({
+                    if(this.sourcewin != null && this.sourcewin !== undefined) {
+                        this.sourcewin.destroy();
+                    }
+
+                    this.sourcewin = new Ext.Window({
                         width:winWidth,
                         id:'processSources',
                         height:winHeight,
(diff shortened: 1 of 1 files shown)


jlindop:
@tsurdilo 
Code looks good and it works well.
+1

jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/view.js
```diff
@@ -61,6 +61,9 @@ ORYX.Plugins.View = {
         this.maxZoomLevel = 2.0;
         this.diff=5; //difference between canvas and view port, s.th. like toolbar??
 
+        // source view window
+        this.sourcewin;
+
```
csadilek: What is this line doing?

tsurdilo: @csadilek its not left-over.  just defines sourcewin which is used later in the methods

csadilek: @tsurdilo it's not needed. you're not defining anything either as it's not being initialized.

this.foo;
alert(this.foo);

this.foo is undefined, of course!

tsurdilo:
@csadilek The JS code gets compressed and minimized so this line if it's there or not does not matter. In the compressed designer-plugins.js it wont be there one way or another. I like it there for readability but if it's some sort of big problem for you it's fine - commit amended.

csadilek:
@tsurdilo I never claimed it was a big problem. We didn't start the code reviews to exclusively spot serious problems but also to improve structural quality. The latter can only be done by reducing technical debt and removing unneeded code is the easiest way to do so.

tsurdilo:
@csadilek ok sounds good. let me know if there is anything else that you see needs to be changed. thanks!

csadilek:
Thanks.

master: 83443a006f6a667bdf91365327c7babccf1f4e1e

PR #103, opened 2015-12-07
[GUVNOR-2278] add missing client dependencies

psiroky:


mswiderski:
+1 as it does not seem to be risky and since it cannot be built at all this commit won't make it worse...

PR #104, opened 2015-12-08
JBPM-4815: added testcases for DataIOEditorNameTextBox and AssignmentListItemWidget

jomarko:


jbpm-designer-client/src/test/java/org/jbpm/designer/client/util/DataIOEditorNameTextBoxTest.java
```diff
@@ -0,0 +1,231 @@
...
+        textBox.setRegExp(ALPHA_NUM_REGEXP, ERROR_MESSAGE_REGEXP);
+
+        INVALID_VALUES.clear();
+        INVALID_VALUES.add("abc");
+        INVALID_VALUES.add("CdE");
+        INVALID_VALUES.add("a#$%1");
+    }
+
+    @Test
+    public void testIsValidValueSensitive() {
```
triceo: This is a lot of code repetition. How about @Parameterized?

jbpm-designer-client/src/test/java/org/jbpm/designer/client/util/DataIOEditorNameTextBoxTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+        isValidResult = textBox.isValidValue("a#$%1", true);
+        assertEquals(ERROR_MESSAGE, isValidResult);
+
+        isValidResult = textBox.isValidValue("a#$%1", false);
+        assertEquals(ERROR_MESSAGE_REGEXP + ": #$%", isValidResult);
+    }
+
+    @Test
+    public void testIsValidValueInsensitive() {
```
triceo: Ditto.

jbpm-designer-client/src/test/java/org/jbpm/designer/client/util/DataIOEditorNameTextBoxTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+        isValidResult = textBox.isValidValue("a#$%1", true);
+        assertEquals(ERROR_MESSAGE, isValidResult);
+
+        isValidResult = textBox.isValidValue("a#$%1", false);
+        assertEquals(ERROR_MESSAGE_REGEXP + ": #$%", isValidResult);
+    }
+
+    @Test
+    public void testMakeValidValueSensitive() {
```
triceo: And again.

jbpm-designer-client/src/test/java/org/jbpm/designer/client/util/DataIOEditorNameTextBoxTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+        makeValidResult = textBox.makeValidValue("c");
+        assertEquals("c", makeValidResult);
+
+        makeValidResult = textBox.makeValidValue("a#b$2%1");
+        assertEquals("ab21", makeValidResult);
+    }
+
+    @Test
+    public void testMakeValidValueInsensitive() {
```
triceo: And one last time. :-)

psiroky:
@jomarko would it be possible to remove the reference to the internal JIRA project RHBXMSQE? I believe this is similar to the BPMSPL where we agreed to not use that in upstream commits.

jomarko:
@psiroky satisfy this PR naming conventions now?

@triceo I didn't manage to combine Parameterized  runner with GwtMockitoTestRunner runner. Do you know how to do it or did you see it anywhere in some existing tests? For now I extracted repeated code to private methods. Is it better?

psiroky:
@jomarko perfect, thanks!

manstis:
@jomarko You can in theory use ```RunWith(Parameterized.class)``` and use ```GwtMockito.initMocks()``` instead of ```RunWith(GwtMockitoTestRunner.class)```. I tried once for too long and went for a "super base test class" with tests sub-classing it. See [here](https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-project-explorer/kie-wb-common-project-explorer-client/src/test/java/org/kie/workbench/common/screens/explorer/client/widgets/navigator/BaseBreadcrumbNavigatorTest.java#L37).

jomarko:
@manstis Great hint! Simple and nice solution. I was trying to combine both annotations and it was enough do not use both annotations :)

I think, PR is ready for review.

jbpm-designer-client/src/main/java/org/jbpm/designer/client/util/AbstractValidatingTextBox.java
```diff
@@ -70,11 +68,13 @@ public void onKeyPress( KeyPressEvent event ) {
                 String newValue = oldValue.substring(0, me.getCursorPos());
                 newValue = newValue + ((char) charCode);
                 newValue = newValue + oldValue.substring(me.getCursorPos() + me.getSelectionLength());
-                if (isValidValue(newValue, false) != null ) {
+                String validationError = isValidValue(newValue, false);
+                if (validationError != null) {
                     event.preventDefault();
+                    fireValidationError(validationError);
```
jlindop: Adding the call to  'fireValidationError' inside onKeyPress works fine, but I think we should use a different error message from the one used when fireValidationError is called from within the BlurHandler. The error message "Removed invalid characters from name: #" makes sense from the BlurHandler, as it's called after the user has pasted an invalid name, e.g. "ab#cd" but from onKeyPress, it would be better to use something like "Invalid character in name: #" (where '#' is the key pressed).
Will you either change this, or creating a  BZ for this change to be made would also be fine. Thanks.

jlindop:
@jomarko  Apart from my comment above about the Name text box validation errors, this is great.
Thanks very much for adding these extra test cases !

jomarko:
Updating according to @jlindop comments

jomarko:
@jlindop there should be used two distinct error messages now.

jlindop:
@jomarko 
That's great! I just tested and it works well.
+1

csadilek:
Thanks!

master: 456a829ee34f2dec93a6f41934eeae4fae44d0b1

PR #106, opened 2015-12-10
Another round of fixes found by findbug.

Hasys:
@jlindop @tsurdilo Can I ask you to review this PR. It already had an conflict so in the future it will be even worse.

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
@@ -1714,9 +1714,9 @@ public class Bpmn2JsonUnmarshaller {
 
         for(BoundaryEvent beEntry : toAddBoundaryEvents.keySet()) {
             BoundaryEvent be = Bpmn2Factory.eINSTANCE.createBoundaryEvent();
-            if(beEntry != null && beEntry instanceof ErrorEventDefinition) {
+            if(beEntry instanceof ErrorEventDefinition) {
                 be.setCancelActivity(true);
-            } else {
+            } else if (beEntry != null) {
                 Iterator<FeatureMap.Entry> iter = beEntry.getAnyAttribute().iterator();
                 while(iter.hasNext()) {
                     FeatureMap.Entry entry2 = iter.next();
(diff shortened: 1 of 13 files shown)


tsurdilo:
Hi, couple comments:
* Bpmn2JsonUnmarshaller - 
changing if(beEntry != null && beEntry instanceof ErrorEventDefinition) { 
to 
 if(beEntry instanceof ErrorEventDefinition) {
and doing the null -check after can cause NPE if beEntry is null

* DesignerInjectionFilter is no longer used to those changes are not needed i think ( same for InjectionConfig and Beautifier)

* IOUtils.closeQuietly - are we sure we want to ignore exceptions here? 

* DesignerInjectionFilterTest - really nice to have but again the injection filter is not used so not sure if we need or not

Hasys:
@tsurdilo thank you for your review.

Bpmn2JsonUnmarshaller:
I tried it locally and no NPE was thrown. Try this example:
public class Test {
    public static void main(String... args) {
        Object newObject = null;
        
        if (newObject instanceof Object) {
            System.out.println("It is always true except null");
        } else {
            System.out.println("It was null");
        }
    }
}

"It was null" was shown no exceptions thrown.

Can you provide example when it can cause NPE?

IOUtils.closeQuietly - yes, why not? It is just closing the stream in finally block.

DesignerInjectionFilter, DesignerInjectionFilterTest and others: so we can remove all this files from repository?

Hasys:
@tsurdilo Update for NPE quote from Java Language Specification:
"At run time, the result of the instanceof operator is true if the value of the RelationalExpression is not null and the reference could be cast to the ReferenceType without raising a ClassCastException. Otherwise the result is false."

tsurdilo:
@Hasys i think thats fine. i think having the null check further up as it was is fine but then do you need it again below? Up to you

Hasys:
@tsurdilo I see.. I will fix that in the further commits. This PR is old enough to postpone it if no regressions or new problems found. I will work with Bpmn2Json[Un]marshaller intensively for nearest future and this thing I will fix firstly.

tsurdilo:
Sounds great. Thank you for all your help. It is really appreciated!

porcelli:
@tsurdilo is this good to merge?

tsurdilo:
@porcelli I think @Hasys meant to incorporate these changes in his new PRs, @Hasys please let us know if I misunderstood something.

Hasys:
@tsurdilo @porcelli you are right new changes will be in new PRs. So this PR is ready for merge.

tsurdilo:
+1 then ;)

porcelli:
https://github.com/droolsjbpm/jbpm-designer/commit/eed5685e6072ecc7165bd1d396d3100013070503

PR #107, opened 2015-12-11
BZ 1003908 - Non-Interrupting boundary event has invalid graphical semantics

tsurdilo:
@jlindop - please test - video on what to look for here: https://streamable.com/jq2e
Note that cancel activity property will only persist if event is a boundary event so if you test events that are not boundary it will revert back to "false" - default. Thanks.

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/stencilsets/bpmn2.0jbpm/stencildata/bpmn2.0jbpm.orig
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/stencilsets/bpmn2.0jbpm/stencildata/bpmn2.0jbpm.orig
@@ -237,21 +237,6 @@
 		{
 			"name":"boundaryevent",
 			"properties":[
-				{
-					"id":"boundarycancelactivity",
-					"type":"Boolean",
-					"title":"CancelActivity",
-					"value":false,
-					"description":"Denotes whether the activity should be cancelled or not, i.e., whether the boundary catch event acts as an error or an escalation. If the activity is not cancelled, multiple instances of that handler can run concurrently.",
-					"description_ja":"アクティビティをキャンセルするかどうか、つまり、Boundaryキャッチイベントがエラーになるかエスカレーションになるかを示します。 アクティビティがキャンセルされない場合、そのハンドラの複数のインスタンスを同時に実行できます。",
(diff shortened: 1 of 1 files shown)


jbpm-designer-client/src/main/resources/org/jbpm/designer/public/stencilsets/bpmn2.0jbpm/stencildata/bpmn2.0jbpm.orig
```diff
@@ -238,19 +238,32 @@
...
-                                                "frame",
-                                                "frame2"
-                                        ]                                               
+						"id":"boundarycancelactivity",
+						"type":"Choice",
+						"title":"Cancel Activity",
+						"value":"false",
+						"description":"Denotes whether the activity should be cancelled or not, i.e., whether the boundary catch event acts as an error or an escalation. If the activity is not cancelled, multiple instances of that handler can run concurrently.",
+						"description_ja":"アクティビティをキャンセルするかどうか、つまり、Boundaryキャッチイベントがエラーになるかエスカレーションになるかを示します。
+アクティビティがキャンセルされない場合、そのハンドラの複数のインスタンスを同時に実行できます。",
```
jlindop: @tsurdilo Somehow a newline has been inserted into the "description_ja" property (I guess this may have happened when you pushed to github).  This breaks the json, so I wasn't able to test the PR as it stands.

jlindop:
I fixed the stencilset locally and tested. Most of the boundary event svg's change correctly, although Error and Compensation don't behave correctly, probably because their svg's have a single circle rather than a double circle around them, which I think is incorrect. See https://vimeo.com/148865792
Looking at the BPMN2.0 spec, p261, I see that Error and Compensation cannot be non-interrupting, so ideally it wouldn't be possible to set the property on these.

tsurdilo:
@jlindop Nice catch. Currently we have cancel activity property in the boundryevents group which will affect all catching events. Will update this.

tsurdilo:
@jlindop amended - cancel activity now local property of the events.

tsurdilo:
amended with default value being true.

jlindop:
The change looks good, and it's all working nicely.
+1

csadilek:
master: 5d3805458af9512fd759acf9d1fccd7dc735bb40

PR #108, opened 2015-12-16
BZ1253617: Size dialogs according to size of designer editor

jlindop:
This fixes the problem where dialog Close / Cancel buttons are sometimes not visible if the browser is quite small.
@tsurdilo Will you take a look at this and test?


tsurdilo:
@jlindop looks good just one question - have you tried to use the browser built-in detection of width / height? for example document.documentElement.­clientHeight/Width instead of writing a function that looks up body element?

tsurdilo:
@jlindop if you are looking at the width/height of the frame itself i think you can get that from the presenters (look at the onresize methods). Just asking if that is something we could try but if it's not then this looks good.

jlindop:
@tsurdilo Thanks for the feedback. I've revised getDialogSize to use document.documentElement.clientHeight / clientWidth

tsurdilo:
@jlindop Thanks! 
+1

csadilek:
master: a04fdb5c35106a58a8dfb226904a13cfefa3d2f9

PR #110, opened 2015-12-17
BZ 1283227 - Read-only mode allows to select elements and see their shape menu

tsurdilo:


tsurdilo:
@jlindop 
I have tested this on kie-wb, if you want to do it also just log in with 2 different users in 2 windows, open the sample evaluation process and lock it (move a mode). then open the same process with another user which will open it in read-only mode. Select a couple of nodes (with click+drag) and the shape menu should not appear.

jlindop:
Tested and works fine.
+1

csadilek:
master: fe23cbd01ce93747e91eba537e7be8c8f555c8c7

PR #111, opened 2015-12-18
BZ 1243597 - Designer does not allow line breaks in many component  - fix for boundary events

tsurdilo:


tsurdilo:
@jlindop 
to test drop a task and a catch event separately. add a multi-line name for the event . move the event to be boundary event of the task. save+re-open. the multi-line name of event should persist

jlindop:
Looks good. Tested and works well.
+1

csadilek:
master: a785cf4a62f33445c46b418366afaa9157305829

PR #113, opened 2015-12-21
BZ1059891: Make sure OK and Cancel buttons on popups are always visible

jlindop:
@tsurdilo Will you take a look at this. I've made sure the popup dialogs launched from the button bar are fully visible even when the browser is quite small, so that the OK and Cancel buttons are always accessible.


tsurdilo:
Looks good
+1

csadilek:
master: ee61cc5bf01b2fc66253b04e93c9841e564a8dcf

PR #116, opened 2016-01-06
bz1289575: Improve error handling in TransformerServlet.doPost in cas…

jlindop:
…e payload is too long

When the user tries to download a PDF or image for a very large process, an error is now reported.

If you're running business-central in EAP standalone, the fix for this is to edit standalone-full.xml, adding the property  max-post-size="0" to
<connector name="http" protocol="HTTP/1.1" scheme="http" socket-binding="http"

@tsurdilo  will you review this - if you can think of a better error message, that would be fine.


csadilek:
Isn't there a way to limit the size of the request so that the issue can be fixed? 

If not and we really just have to throw an error then we still need to add test for this change! 

What's the purpose of logging an "invalid request" error and then still keep going and executing the rest of the request processing logic?

jlindop:
Closing as there is a way to reduce the request size. Will create a new PR for this.

PR #122, opened 2016-01-12
Fix tests failing on Java 8 + small test refactoring

psiroky:
* the order of elements in the JSON string is not defined
   and on Java 8 it is different that before. The tests
   are now checking separately for all specific parts of
   the string, instead of the string as a whole

@tsurdilo, @jlindop could you please take a look?


jlindop:
@psiroky Thanks for fixing these unit tests
+1

csadilek:
master: f7761cc95949413da943e75dfb73a53a413d8655

PR #126, opened 2016-01-21
BZ1248991 show Input/Output counts on Assignments properties

jlindop:
@tsurdilo Will you try this. The value displayed in the UI for Assignments has been replaced with a count of the numbers of Inputs and Outputs, with a note if there some are incomplete, i.e. missing either data-type or a source/target. This UI idea was a suggestion from Joe.


tsurdilo:
Hi Jeremy, this is a big change :)
I tested and found 2 issues:

1. Does not work for Service Tasks: http://imgur.com/WU2Tp1X
2. Does not work for empty assignments to a constant (for example data input x of type String assigned to "") : http://imgur.com/IfTrYU3 (still says its incomplete)

Also since this is a new feature I would suggest to not use "inputs" and "outputs"  but maybe "datainputs" and "dataoutputs" so its not confusing the users with the actual sequenceflow connections to and from the task node. WDYT?

jlindop:
@tsurdilo Thanks for the feedback. I've updated the PR to fix the problem with service tasks and empty assignments. I've removed the 'incomplete' count, because, as you say, this could be confusing. I also changed the wording from inputs / outputs to  data inputs / data outputs.

tsurdilo:
@jlindop looks good. thanks! I added  https://github.com/droolsjbpm/jbpm-designer/commit/3ecdf67150a7c7681ca635dc474a196e10aa5bf2 in that branch to compile it so can test.

+1

csadilek:
master: d1661036efe746a1191b50aa662e35c6bfb06635

PR #127, opened 2016-01-21
JBPM-4908  Add zanata.xml config file to jbpm-designer-api module

jlindop:
Added after discussion with Michael Biarnes Kiefer.

### DIFF
--- a/jbpm-designer-api/src/main/config/zanata.xml
+++ b/jbpm-designer-api/src/main/config/zanata.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<config xmlns="http://zanata.org/namespace/config/">
+  <url>https://translate.jboss.org/</url>
+  <project>jbpm-designer</project>
+  <project-version>6.4.0</project-version>
+  <project-type>properties</project-type>
+
+  <locales>
+      <locale map-from="zh_CN">zh-Hans</locale>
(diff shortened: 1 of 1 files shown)


manstis:
Is there a very good reason why there's i18n in the -api module  (I assume possibly i18n for server - side generated messages shown on the client?).. otherwise it's questionable why an -api module needs Zanata configuration.

jlindop:
@manstis Yes, designer has a set of messages in the api module which are shown on the client.  The zanata.xml file was added to the 6.3.x branch, and I guess the fact that it wasn't also added on master was accidental.

manstis:
Fair enough, I appreciate that. 

I was just a little surprised that an -api module needs Zanata configuration and perhaps it highlights that the i18n constants would be better stored in the -client (or -backend module - I suspect Designer includes the i18n constants in a download for Oryx from a Servlet or similar)...

mbiarnes:
I did another test, like bespoken with Neus.
The steps to do Zanat translation updates in jbpm-designer are:
in root directory mvn zanata:pull-module; mvn replacer:replace -N
then cd into jbpm-designer-api and do again replacer:replace -N.
This works without this zanata.xml.
Before we applied the ascii2ascii and did in jbpm-designer-api a mvn zanata:pull-module again, therefor we needed the xml - but the preocedure changed a bit.
I think this PR can be closed and not merged in.
Sorry for the caused inconveniences.

jlindop:
@mbiarnes - Closing without merging, as suggested.

PR #137, opened 2016-01-29
Increase memory for GWT compiler to 2g in full profile

psiroky:
Needed to fix a failing build, as the original 1g is not enough anymore.


jlindop:
I found that <extraJvmArgs>-Xms512m -Xmx2048m -Xss1M</extraJvmArgs>  was sufficient, but more is fine.
+1

PR #138, opened 2016-01-29
Increase -Xmx Jvm parameter to enable -Dfull compilation

jlindop:
Compilation of jbpm-designer-standalone with -Dfull option was failing with Out of Memory error.


jlindop:
Closing, as @psiroky just pushed a PR for the same problem a few minutes ago!

PR #148, opened 2016-02-12
[GSS-RFE]Export business process to PDF does not includes documentation properties in BPM Suite 6

tsurdilo:
@jlindop this includes the docs and your code for the pdf generation


jlindop:
Tested and works well.

+1

jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/documentation/docstructure.js
```diff
@@ -0,0 +1,659 @@
...
+        } else {
+            return propvalue;
+        }
+    } else {
+        return propvalue;
+    }
+}
+
+function presentPropertyName(rawprop) {
+    if(rawprop == "assignmentsview") {
```
csadilek: These are all static values so why not simply replace the giant if/else if with a associative array/map lookup i.e. define somehwere ```var propertyNames = {"assignmentsview": "Assignments Count", "script_language": "Script Language", ...}``` and then this method can simply look up using return ```propertyNames[rawprop]```.

jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/documentation/docstructure.js
```diff
@@ -0,0 +1,659 @@
...
+                var innerParts = nextPart.split("=");
+                if(innerParts[0].startsWith("[din]")) {
+                    var fromPart = innerParts[0].slice(5, innerParts[0].length);
+                    innerParts.shift(); // removes the first item from the array
+                    var escapedp = innerParts.join('=').replace(/\#\#/g , ",");
+                    escapedp = escapedp.replace(/\|\|/g , "=");
+
+                    retVal += "(Data Input) " + fromPart + " is given value " + escapedp + ", ";
+                } else if(innerParts[0].startsWith("[dout]")) {
+                    var fromPart = innerParts[0].slice(6, innerParts[0].length);
```
csadilek: This code is duplicated and needs to be refactored (factored out in a separate method). The only difference seems to be the first param to slice.

jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/documentation/docstructure.js
```diff
@@ -0,0 +1,659 @@
...
+            } else {
+                // default to equality
+                if(innerParts[0].startsWith("[din]")) {
+                    var fromPart = innerParts[0].slice(5, innerParts[0].length);
+
+                    retVal += "(Data Input) " + fromPart + " is given value  , ";
+                } else if(innerParts[0].startsWith("[dout]")) {
+                    var fromPart = innerParts[0].slice(5, innerParts[0].length);
+
+                    retVal += "(Data Output) " + fromPart + " is given value  , ";
```
csadilek: if and else if block are exactly the same! This is either wrong and needs to be refactored.

csadilek:
@tsurdilo three minor issues above. Once fixed/tested, we can merge, given that it's all JS and the designer gets special treatment for not requiring tests there.

tsurdilo:
@csadilek thanks im on it

tsurdilo:
@csadilek I have amended updates to your comments 2 and 3. I will look into refactoring the code for your comment 1 as soon as possible.

csadilek:
Build failure unrelated.

csadilek:
master: 49aafe71a8571237037c2c2efde1a3adc4b74732

PR #151, opened 2016-02-16
BPMSPL-230 - added process image for view and print versions. updated  pdf file name and other small fixes

tsurdilo:
@jlindop some updates to process docs:
1. Added process image as item 3.0 to the docs. There are 2 versions of this process image one for the view inside designer which has a black border and scroll bars to navigate larger processes. If user selects "Print" and prints to PDF or via printer the image is not inside a bordered div and has a much larger display to fit larger processes.
2. When saving the doc PDF the filename of the pdf contains the process asset name, for example evaludationDocumentation.pdf
3. The Navigation bar and the area with the PDF and Print buttons is no longer visible when user selected Print and prints to PDF or via printer

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/documentation/docstructure.js
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/documentation/docstructure.js
@@ -5,7 +5,7 @@ var elementDataInfo = {
         "dispname": "User Task",
         "groupicon": "stencilsets/bpmn2.0jbpm/icons/activity/task.png",
         "icon": "stencilsets/bpmn2.0jbpm/icons/swimlane/process.participant.png",
-        "properties": "name,documentation,datainputset,dataoutputset,assignments,isasync,taskname,actors,groupid,subject,description,skippable"
+        "properties": "name,documentation,datainputset,dataoutputset,assignments,isasync,taskname,actors,groupid,subject,description,skippable,content,createdby,locale,multipleinstance,notifications,priority,reassignment"
     },
     "SendTask": {
         "group": "Tasks",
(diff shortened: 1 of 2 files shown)


jlindop:
@tsurdilo The pdf's still don't work very well - I'll show you later today.
About item 3.: When I use the "PDF" button, the PDF still contains the nav bar and buttons

tsurdilo:
@jlindop yes I didn't change the PDF functionality ye (the pdf that we generate)t. Will do a separate PR for that.
regarding 3: Sorry I did not write this well enough - click on "Print" button then use browser to save or view it as PDF. The navbar and the area with buttons will not be on the PDF (see example http://imgur.com/UV64BZf)

tsurdilo:
@jlindop amended small change to allow pdf (generated by browser via print button) to not look squished like before.

tsurdilo:
@krisv @jlindop 
we need to look further at the PDF generation as what we have currently is not good enough to be in the product (stretching of the html content occurs which makes the generated PDF not readable). I have looked at iText however it does require further libraries which are not included in any of our poms nor the jbpm nexus to deal with css and formatting of css and will investigate further on this.

I have removed temporarily the PDF generation and have replaced it with much better PNG views of the process documentation and the process image (separate buttons).

tsurdilo:
@krisv @jlindop -- changed the display of the "totals" to simple lists as QA noticed they were not aligning correctly at times. Also updated all element display properties according to the QA doc

tsurdilo:
updated the display for data inputs, data outputs and assignments. now each entry will be displayed on its own line and no longer as one long string value (for readability and also mentioned as an issue in the QA doc)

tsurdilo:
@krisv @jlindop 
To summarize changes in this PR:
1. Temporarily removed "PDF" button while we are working on this
2. Added "Documentation PNG" and "Model PNG" buttons 
3. Updated all element properties that are displayed according to the QA findings
4. Updated display of "Totals" for both process and elements
5. Fixed bug with properties that are boolean type
6. Updated display for data inputs/outputs, assignments, and vardefs (each is displayed in its own line now)

jlindop:
@tsurdilo The "Model PNG" button is working on Firefox, but not on Chrome. I get  a small image with just the text "Eval"

tsurdilo:
@jlindop yes chrome is being difficult with base64 encoded images. Model PNG button removed because users can get it from the menu bar anyways. Amended

jlindop:
Looks good now.
+1

psiroky:
We should not reference non-public BPMSPL JIRAs in the commit messages. @tsurdilo could you please fix the commit message to reference JBPM one? Thanks!

tsurdilo:
@psiroky done

psiroky:
@tsurdilo thanks, I will merge this one the build succeeds.

PR #155, opened 2016-02-19
BZ1241550: Fix json to bpmn2 unmarshalling exception for ServiceTask …

jlindop:
…with no params


tsurdilo:
@jlindop I think the real bug is the condition under which the assignments property in the json becomes just "="

"assignments": "=",

This could have been introduced by the number of refactoring prs lately but maybe worth investigating because it should imo never be just "="

tsurdilo:
@jlindop this seems to be in bpmn2.0jbpm.orig template where stringtemplate parses the workitems:

"value":"$workitemDefs.(k).parameters:{k1| $k1.name$}; separator="=,"$=",

jlindop:
@tsurdilo  Thanks - your idea to fix the root of the problem in bpmn2.0jbpm.orig template works, and I've updated the PR accordingly. There was no need to change the value for the results of workitems in the template as the value of dataoutputset for services with no results is already "".

tsurdilo:
+1

PR #157, opened 2016-02-23
BZ1311039 - Some designer tests are not running on Jenkins 

hasys:
Parent surefire plugins settings run only "*Test.java". Class renamed, some tests fixed (another 3 seems like real bugs.)


jlindop:
@tsurdilo The tests which now fail are testPoolUnmarshalling, testDataStoreUnmarshalling and testMessageUnmarshalling which run with json processes containing references to stencils with id's "Pool", "DataStore" and "Message" respectively. Are these stencils which have been removed from the stencilset? If they are, can the tests and their association json resources be deleted?

tsurdilo:
@jlindop yes I think that would be fine

Hasys:
@jlindop @tsurdilo It seem like "Pool" is a name of a process not a stencil's id. Am I right? But process after unmarshalling have empty name property.

jlindop:
@Hasys In the test file "pool.json", there's a reference to a stencil called "Pool" in the json:
        "stencil": {
            "id": "Pool"
        },
The current stencilset doesn't have a stencil with id "Pool", so I think this test is not valid any more.
The same is true for the "DataStore" and "Message" tests which also fail, so I think it makes sense to remove the tests and their associated json files. It would be great if you'd update the PR to do this.

Hasys:
@jlindop sorry, but I little bit misunderstood. 
It fail on
`assertEquals("pool", process.getName());`
if you will comment this line test will pass, but resulted XML have no process name.
Also test "testLaneUnmarshalling" use "pool.json" and passes.
As I understand Unmarshaller  parse "processn" for Name property of Process, but in test we have "name" and "processname" properties. Is it OK? When I changed pool.json ("name": "pool" -> "processn": "pool") it passed.

jlindop:
@Hasys OK, sorry, I misunderstood. Yes, I think changing pool.json in the way you suggest is a good idea.

hasys:
@jlindop Pool test is fixed now, another two was removed, because tested outdated and removed functionality.

jlindop:
@hasys Thanks for fixing these.
+1

PR #158, opened 2016-02-23
GUVNOR-2415: Boundary events lose it’s documentation after save and r…

Hasys:
…eopen Business Process + fixed missing/wrong named resources


tsurdilo:
+1 Thanks!

PR #164, opened 2016-03-01
JBPM-5011: Artifact (Group) lose documentation

hasys:


jlindop:
@hasys The fix is good. I'm not so sure about the test case in Bpmn2JsonMarshallerTest, which starts from json rather than bpmn2. For now, I think it makes sense to create a separate BZ to add a set of test cases in Bpmn2JsonMarshallerTest, then address the test in that BZ.
+1

csadilek:
6.4.x: 35cef1d035b81c943a052fe5ed2c89cf4560cd05

PR #165, opened 2016-03-01
JBPM-5011: Artifact (Group) lose documentation

hasys:


jlindop:
@hasys My comment about the test in PR164 also applies here. It probably makes sense to fix and extend the test cases via a separate BZ.
+1

csadilek:
master: 55ed11a0d46586f2982a71164f7fb83e4d92df18

PR #166, opened 2016-03-02
JBPM-4999: Fix IE11 problem caused by Javascript String not implement…

jlindop:
…ing startsWith function


csadilek:
+1

csadilek:
6.4.x: 680035857d63e6b8c9712cb3fa8d37056bb528b6

PR #168, opened 2016-03-03
JBPM-4999: Remove process image from Process Docs tab because it does…

jlindop:
…n't work in Chrome and it breaks designer in IE11.

@tsurdilo I've created this PR as a solution that works. Feel free to reject it if you think there's a better way (-:


tsurdilo:
+1 since urgency and we should look into making this work soon

csadilek:
6.4.x: 984c03aa4ceadb98e522d60f73e204a14c96ca0e

PR #175, opened 2016-03-04
BZ 1314510 - process model image shows the node in waiting status grey instead of red after sending a signal event

tsurdilo:
@jlindop this is how to test (you must test this inside the workbench, not designer standalone)

1. create new process and use following bpmn2 for test: http://pastebin.com/u6h6FMGB
2. Save process and generate all forms (important)
3. Build and deploy project
4. Go to process management and start instance of this process
5. on the process start form select the "loop" checkbox
6. Signal the process (signal name is "test")
7. View the process instance view (designer in readonly mode)
8. The signal node should be highlighted as active (See http://imgur.com/E5tA82P)

Before this change the signal node after the signal which triggers the loop was grey and not "active"


jlindop:
Tested and works fine.
+1

csadilek:
master: 9e139fdb0a7af958eeb160c55f5d4684819a42e1

PR #183, opened 2016-03-10
JBPM-5013_4.x Ensure Assignments with names e.g. "Content", "Locale" …

jlindop:
…etc on non-User tasks are saved.


jlindop:
Closing - will fix commit message.

PR #185, opened 2016-03-10
BZ 1315981 - Process preview in process instances not loaded when opened for 2nd time

tsurdilo:


manstis:
How does this solve the issue? Why would the event interfere with opening a read-only process for the second time? Test covers change... but does the "fix" solve a symptom of a more deeply nested issue or does it solve the cause?

tsurdilo:
Yes it solves the problem. The exception does nkt happen in designer which i debugged but in the workbench. I assume there is code which onserves this event and tries to use the file system

tsurdilo:
Not sure that was the point of what i said. I did not debug the workbench because there is nothing wrong with dealong with vfs when this even is onserved so its not a bug. It is that when dealing with designer in read only mode the vfs does not exist so the fix in designer is fine

tsurdilo:
There is no reason for the event to be raised in designer read only mode tobegin with

tsurdilo:
@manstis nooooo i reserve the right to bug you about all events:) no really thank you for always helping me i really appreciate it

csadilek:
master: f2b2c8987787a9f97e6209bd0ca4ab76870a92a2

PR #189, opened 2016-03-14
JBPM-5007: stop showProcessDocs function from being called multiple t…

jlindop:
…imes when not needed. This was impacting performance noticeably in IE11.

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Core/main.js
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Core/main.js
@@ -437,6 +437,9 @@ ORYX.Editor = {
 		        deferredRender : false,
 		        listeners: {
 	             	tabchange: function(tabpanel, tab) {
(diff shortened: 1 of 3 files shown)


csadilek:
Needs a review. @tsurdilo, maybe?

tsurdilo:
@jlindop any reason you changed 

document.getElementById("processdocframe").contentWindow.showProcessDocs();
to
this.processDocContentPanel.items.get('processdocframe').getEl().dom.contentWindow.showProcessDocs();

your approach uses extjs lib which maybe is better to start avoiding if possibl

other than this  +1

jlindop:
@tsurdilo I was getting exceptions when developing this at first, hence the extra 'if ()' clauses and long-winded way of calling showProcessDocs(). It's working now with the simplified call to showProcessDocs(), so that's a good catch (:

tsurdilo:
+1

csadilek:
master: 3d53541e6686c3db5e49bd70c467bc137bcda6f7

PR #191, opened 2016-03-15
BZ1297383: Add style property with height and width to PDF created by…

jlindop:
… "Share Process PDF" menu option


csadilek:
Needs a review, @tsurdilo?

tsurdilo:
@jlindop I think finding the SVG width height by traversing the svg in backend is really overhead when you already get it for free from svg element returned from getSVGRepresentation in canvas.js

Once you get the svg representation you should be able to simply say:

getCanvas().getSVGRepresentation(false).getAttributeNS(null, 'width')
getCanvas().getSVGRepresentation(false).getAttributeNS(null, 'heght')

and pass it into transformer servlet. IMO this would be a better approach and more stable. WDYT?

tsurdilo:
+1

csadilek:
master: 3db10a970d28b2fab32dac7da314ac8203216d65

PR #197, opened 2016-03-18
JBPM-5035 Replaced System.lineSeparator method call

baldimir:
- This is relevant only for 6.4.x branch, because on master it's expected that it supports jdk 1.8 and up.


jlindop:
+1  (for JDK 1.6 compatibility)

csadilek:
6.4.x: ea4dc1408da60e92c15c7288a4b913f5d4ac7bf2

PR #198, opened 2016-03-21
JBPM-5007 - Process Documentation not contain Process Image

tsurdilo:


tsurdilo:
@jlindop what you showed me in your local branch did not work for me on Chrome. I changed the code to use the max size of the canvas thats even possible on the modelling side and tested with number of large processes. worked fine on chrome and ff as well this way (amended)

jlindop:
@tsurdilo This works fine in IE11 now, as well as FF and Chrome.
+1

csadilek:
master: cffbdff843bf67772ed96579d771d07ce2225337

PR #204, opened 2016-03-21

tsurdilo:
… new line and tabulation


jlindop:
@tsurdilo This fixes the cases in the JIRA entry, but doesn't work correctly if a string constant contains "\n", e.g. System.out.println("Hello\nGoodbye");

tsurdilo:
@jlindop thanks - the method from propertywindow (and revised one you send) had some issues so I amended with a small change that seems to fix most cases. If you get that method to do what we need do feel free to change this to it

tsurdilo:
@jlindop amended

PR #205, opened 2016-03-21

tsurdilo:
… new line and tabulation


tsurdilo:
@jlindop amended

PR #206, opened 2016-03-22
JBPM-4998 Hide DataInputSet and DataOutputSet for service tasks which…

jlindop:
… don't use customEditor.

@tsurdilo  This fixes the problem reported in JBPM-4998. I didn't add  '"customEditor" : "true"' back into WorkDefinitions.wid.st  because if services have results as well as parameters, these can't be edited when customEditor is true.


tsurdilo:
@jlindop ok idk if just removing it again fixes any problems as some users might have customEditors flag in their wid. I'll take a look at wid with results array and fix problems first then we could look at this again.

tsurdilo:
@jlindop when users chose customEditor in the wid it makes no sense to show results mapping but only the parameters which it happens currently as expected (see http://imgur.com/9DXlUpV). I think if users want to use the property-per parameter option then we should let them instead of just disabling it (again :) )  as it's in this PR. 

I am not sure why the Jira says they see different editors but it could be that https://github.com/droolsjbpm/jbpm-designer/commit/67cacebe1b8ab46711190c9e93291e08ed4c2054 was not added to the proper branch which now should no longer be the case as it seems to be in 6.4.x.

tsurdilo:
@jlindop my last comment was incorrect (as you showed me). Sorry.

+1

csadilek:
master: 611c5cd8ec8bdaed0078106c6ed0c7d70f01cf01

PR #209, opened 2016-03-23

tsurdilo:
… new line and tabulation

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/documentation/docstructure.js
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/documentation/docstructure.js
@@ -645,6 +645,8 @@ function parsePropertyValue(propname, propvalue) {
         }
 
         return retVal;
+    } else if(propname == "script") {
+        return formatScript(propvalue);
     } else {
         return propvalue;
     }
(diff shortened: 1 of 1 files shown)


jlindop:
Looks good.
+1

csadilek:
Any reason you didn't use a replaceAll function [1] here? 

Seems it would require significantly less code and you'd get a reusable method for other replacements in the future.
replaceAll(script, "\\n". "<br>") (and, btw, br is a void element and doesn't need closing).

[1] http://stackoverflow.com/questions/1144783/replacing-all-occurrences-of-a-string-in-javascript (see second comment)

tsurdilo:
@csadilek we are trying to display the script entered as users will see in the bpmn2 source. we use a very similar method for the script editor as well. In the case of let's say a script 
System.out.println("Hello\nThere");\n we would want to only replace the last \n and not the one user entered themselves so replaceall would not work in all cases here.

csadilek:
@tsurdilo but your formatScript method does go through the whole string and replaces all occurrences of "\n". So, in your example (System.out.println("hello\nthere");\nSystem.out.println("\thow\n\nare\tyou");\n") it replaces all ```\n```s with ```<br>```s. It behaves exactly like ```replaceAll``` would.

tsurdilo:
@csadilek i'm not sure thats true see:
script editor: http://imgur.com/l6RPPF1
documentation script entry: http://imgur.com/S5qhWH9
string that is generated by script task script property editor: System.out.println("Sending Proposal to: "+mail);\nSystem.out.println(" \\t Acme Inc. offers you: "+offering);\nSystem.out.println("hello\\nthere");\nString s1 = "ab\\tcd\\nef";\nString s2 = "ab\\tcd\\nef";\n\nSystem.out.println("hello\\nthere");\nSystem.out.println("\\thow\\n\\nare\\tyou");

csadilek:
@tsurdilo I just ran your function in javascript and provided your example input. It had exactly the same output as replaceAll. Also, looking at the implementation, can you explain how this is not replacing all occurrences of i.e. "\n". It clearly goes through the whole provided string and replaces them all, no?

tsurdilo:
@csadilek I think you forget that the string that we are parsing is not a string as you might think like in our examples here but it comes from the JSON representation of the process. As such what users enter in the script editor has to be escaped to be able to be placed into the JSON and THAT is what we are parsing. 

Here is the JSFiddle for this: https://jsfiddle.net/o5ktkmee/24/

I hope this clears it up, if not let me know

csadilek:
@tsurdilo yes that was the part I didn't know, that the input is already double escaped! However, you can still use ```replaceAll``` for this as it accepts a regular expression. I've updated your jsFiddle to show how (https://jsfiddle.net/o5ktkmee/28/). I will merge this anyway as I've bugged you enough and you can always add a refactoring later.

csadilek:
6.4.x: 83a34f90a1a1276e81c09b26f21d209dda42b3af

jlindop:
@csadilek, @tsurdilo Sorry to labour the point given that it's now been committed, but the regexp doesn't handle tabs and new-lines outside of string constants correctly, whereas the function does. In https://jsfiddle.net/3ge9y05b/  I've added some newlines and a tab between the two lines of code to show this. Maybe this can also be done in a regexp, I don't know (-:

csadilek:
@jlindop sure, we'd need to enhance the regexes. I just made the one example work. Anyways, it's up to you. I just think that having a replaceAll gives us a reusable method and requires less code overall. We already have enough undocumented and untested code to maintain ;).

PR #211, opened 2016-03-24
GUVNOR-2340 - Annotation for element inside subprocess fails validation

tsurdilo:


tsurdilo:
@csadilek @manstis amended

manstis:
+1 to merge.

csadilek:
6.4.x: ef99170222221202e449b0dd2fb5f9ca2c0cb7a9

psiroky:
I've noticed just now (yeah too late :)), but the code uses `tab`s for indentation. We should only use spaces.

PR #212, opened 2016-03-29
JBPM-4979 - Generated PDF is unreadable

tsurdilo:


tsurdilo:
there is a related change to the droolsjbpm-build-bootstrap pom that needs to get merged and propagate before this will pass build

tsurdilo:
Need to mavenize the depends of the used library and will do another PR

PR #232, opened 2016-04-13
BZ1269969: Fix restore BP functionality

jlindop:
The use-cases in Comments 10 & 11 of https://bugzilla.redhat.com/show_bug.cgi?id=1269969 should both not be reproducible any more. (On master, the use-case in comment 10 worked even before this PR).
The case in Comment 11 cannot be reproduced any more because when the user selects a previous version in the Overview tab, the version is opened in read-only mode (whereas previously it was editable).

This PR also fixes the problem where if you select a previous version of a BP, then hit "Restore", the "Restore" button didn't change back to "Save".

@Tiho - testing this PR should include testing that the locking functionality works when two different users have the same BP open.

### DIFF
--- a/jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerPresenter.java
+++ b/jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerPresenter.java
@@ -598,7 +598,7 @@ public class DesignerPresenter
 
     }
 
-    private void setup( Map<String, String> editorParameters,
+    protected void setup( Map<String, String> editorParameters,
                         String editorID,
                         Overview overview ) {
 
@@ -608,9 +608,7 @@ public class DesignerPresenter
 
             resetEditorPages( overview );
 
-            if ( editorParameters.containsKey( "readonly" ) ) {
-                isReadOnly = Boolean.valueOf( editorParameters.get( "readonly" ) );
-            }
+            editorParameters.put("readonly", Boolean.toString(isReadOnly));
 
             designerEditorParametersPublisher.publish( editorParameters );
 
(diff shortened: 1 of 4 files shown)


tsurdilo:
@jlindop can you also run this by @csadilek as he is the locking / unlocking master (share screen or send him video). Also would be nice to add test for DesignerPresenter setup for your line
editorParameters.put("readonly", Boolean.toString(isReadOnly));
WDYT?

jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerPresenter.java
```diff
@@ -608,9 +608,7 @@ private void setup( Map<String, String> editorParameters,
 
             resetEditorPages( overview );
 
-            if ( editorParameters.containsKey( "readonly" ) ) {
-                isReadOnly = Boolean.valueOf( editorParameters.get( "readonly" ) );
-            }
+            editorParameters.put("readonly", Boolean.toString(isReadOnly));
```
tsurdilo: Where does isReadOnly come from ?

csadilek:
Changes look good from a locking perspective. Of course, still needs testing and a plus one before it can be merged.

tsurdilo:
@csadilek ok thanks
@jlindop tested this with two users as well and it looks good. Just wanted to know if adding a test would be good or not or we can do that later too.

+1 (unless Jeremy wants to amend a test)

jlindop:
I'm in the middle of adding a test...

jlindop:
Test now added.

tsurdilo:
Cool.
+1

PR #238, opened 2016-04-20
BZ-1268754: dockers of edges inside of swimlanes are moving now

jomarko:


jlindop:
I've tested this and seen Tiho's comment at https://bugzilla.redhat.com/show_bug.cgi?id=1268754#c3 
Thanks
+1

PR #239, opened 2016-04-20
Bpmn2JsonMarshallerTest - fix test compilation on JDK 1.6.

baldimir:
When trying to run test Bpmn2JsonMarshallerTest on jdk 1.6, the test doesn't compile. I didn't create JIRA for this, because it is only a minor test fix. 

- Fixed usage of method from Files available only on JDK 7+ - replaced with method from Apache IO utils.


psiroky:
+1 to merge (once the build succeeds)

PR #227, opened 2016-04-11
(6.4.x) JBPM-5060: removed unused private method

jomarko:

PR #241, opened 2016-04-20
BZ 1318344 - BPMN Process with swimlane created from business-centra has been distorted while opening in JBDS

tsurdilo:
@jlindop running the test is enough for the test (no need to check in eclipse unless you feel like it)

thanks.


jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2UnmarshallingTest.java
```diff
@@ -1016,4 +1020,25 @@ public void testDocumentationForSwimlane() throws Exception {
         assertEquals("Swimlane name is wrong.", lane.getName(), "Documented Swimlane");
         assertEquals("<![CDATA[Some documentation for swimlane.]]>", lane.getDocumentation().get(0).getText());
     }
+
+    @Test
+    public void testSequenceFlowPointsInsideLane() throws Exception {
+        Bpmn2JsonUnmarshaller unmarshaller = new Bpmn2JsonUnmarshaller();
```
jomarko: Hello @tsurdilo , yesterday there was merged my commit which simplifies initialization of each testcase. Definitions can be now loaded from json in the following way:
`Definitions definitions = loader.loadProcessFromJson("sequenceFlowPointsInsideLane.json");`

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
```diff
@@ -246,6 +246,32 @@ public void revisitDI(Definitions def) {
         revisitEdgeBoundsInContainers(def);
     }
 
+    public Bounds getShapeBoundsForElement(BaseElement fn, BPMNPlane plane) {
```
jomarko: Please do not add this method. Such method already exists on master branch:

https://github.com/droolsjbpm/jbpm-designer/blob/master/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java#L583

Feel free to rename the existing method above if you feel that it doesn't describe itself enough.

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
```diff
@@ -2911,29 +2954,30 @@ private void createSubProcessDiagram(BPMNPlane plane, FlowElement flowElement, B
...
-				}
-				point = dcFactory.createPoint();
-				if(sequenceFlow.getTargetRef() != null) {
-					Bounds targetBounds = _bounds.get(sequenceFlow.getTargetRef().getId());
-					point.setX(targetBounds.getX() + (targetBounds.getWidth()/2));
-					point.setY(targetBounds.getY() + (targetBounds.getHeight()/2));
-				}
-				edge.getWaypoint().add(point);
-				plane.getPlaneElement().add(edge);
+                SequenceFlow sequenceFlow = (SequenceFlow) subProcessFlowElement;
```
jomarko: For better readability I propose to align this block of code in standard way:
`
if (...) {
    // any code here
} else if (...) {
   // any code here
}
`

jomarko: For better readability I propose to align this block of code in standard way. For example:
https://github.com/droolsjbpm/jbpm-designer/blob/master/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java#L2954-2977

tsurdilo:
@jomarko Thanks for the suggestions. The problem I have with this however is that the master branch currently does not run on wildfly and the way we have to do is test with 6.4.x and then cherry-pick to master. Your suggestions are not in 6.4.x so this stuff can happen. 

If I make these changes I will not be able to re-test this vs master currently so I guess we merge this as is (as its tested and your suggestions are something we can revise later as well) or don't merge at all until I am able to run master on 6.4.x . 
@krisv any suggestions?

psiroky:
On master we will support only WildFly 10+. I would not invest the time to make it run in WildFly 8, but rather in WildFly 10.

jomarko:
@tsurdilo yes, my changes are not on 6.4.x branch because @csadilek announced me in #227 that 6.4.x branch is in blockers only mode. So after #227 I stopped creating PR's for 6.4.x branch if I wanted contribute only with refactoring or adding tests. I thought that blockers only mode means that only fixes for BZ's with flag blocker can be merged into 6.4.x branch. Maybe I understand blockers only mode bad.

tsurdilo:
@jomarko no problem. i will run on wildfly 10+ and update this pr . thanks @psiroky !

tsurdilo:
i will redo this PR

PR #242, opened 2016-04-20
JBPM-4979 - Generated PDF is unreadable - update: pdf generation enabled via system property

tsurdilo:
in the process documentation tab the "PDF" button is now disabled by default. To enabled it use startup parameter:

./standalone.sh -Ddesigner.showpdfdoc=true

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/server/EditorHandler.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/server/EditorHandler.java
@@ -85,6 +85,8 @@ public class EditorHandler extends HttpServlet {
      */
     public static final String USEOLDDATAASSIGNMENTS = "designer.useolddataassignments";
 
+    public static final String SHOW_PDF_DOC = "designer.showpdfdoc";
+
     public static final String PRESET_PERSPECTIVE = "org.jbpm.designer.perspective";
 
     /**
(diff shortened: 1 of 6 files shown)


jbpm-designer-backend/src/main/java/org/jbpm/designer/server/EditorHandler.java
```diff
@@ -85,6 +85,8 @@
      */
     public static final String USEOLDDATAASSIGNMENTS = "designer.useolddataassignments";
 
+    public static final String SHOW_PDF_DOC = "designer.showpdfdoc";
```
etirelli: is this the standard for property naming that you use for designer? I see below org.jbpm.designer.XXX and above designer.XXX. Ideally all properties should follow the same convention.

krisv:
+1

csadilek:
+1, I'd prefer a different method name e.g. (is)PdfGenerationEnabled to indicate the boolean return type. Minor thing but if the property name is updated it makes sense to improve the method name as well.

jbpm-designer-backend/src/main/java/org/jbpm/designer/server/EditorHandler.java
tsurdilo: For startup system properties we do "designer.xyz". I will go over all of them to make sure they follow same standard

tsurdilo:
@csadilek agreed. i will make a revision of the entire class and include this

tsurdilo:
@manstis I was not going to make a revision of all properties at a later point (e.g next week) . not within this PR. Maybe that is what is requested but my idea was to do this at a later point.

PR #243, opened 2016-04-20
JBPM-4979 - Generated PDF is unreadable - update: pdf generation enabled via system property

tsurdilo:
in the process documentation tab the "PDF" button is now disabled by default. To enabled it use startup parameter:

./standalone.sh -Ddesigner.showpdfdoc=true

6.4.x commit


krisv:
+1

manstis:
See https://github.com/droolsjbpm/jbpm-designer/pull/242.

I assume changes made there will also be reflected here?

tsurdilo:
@manstis I was not going to make a revision of all properties at a later point (e.g next week) . not within this PR. Maybe that is what is requested but my idea was to do this at a later point.

PR #246, opened 2016-04-25
BZ 1318344 - BPMN Process with swimlane created from business-central as been distorted while opening in JBDS

tsurdilo:
@jlindop  master PR - running test enough. thanks


jlindop:
+1

PR #247, opened 2016-04-25
BZ 1318344 - BPMN Process with swimlane created from business-centra has been distorted while opening in JBDS

tsurdilo:
6.3.x PR for this BZ


jlindop:
+1

PR #248, opened 2016-04-26
JBPM-5091: tests added and some refactoring done

jomarko:
This should be fix for:
https://bugzilla.redhat.com/show_bug.cgi?id=1086912


jlindop:
Looks good.
+1

tsurdilo:
@jomarko I have put https://bugzilla.redhat.com/show_bug.cgi?id=1086912 in "needinfo" with explanation and request to show if the current behavior actually causes errors or not. Let's please wait on the outcome of the BZ before we add any fixes for it. Thanks.

jomarko:
This PR will be amended, please do not merge yet.

jomarko:
Hello @tsurdilo, I have read your conversation with reporter of BZ#1086912. I agree that it would be better fix as UI part so sim engine part of the issue. However I think that this commit fix exactly the issue that was reported and could be merged because as the age of BZ#1086912 suggests,  there will be probably no time to implement changes also to sim engine because of other issues with higher priority.

tsurdilo:
@jomarko ok as we talked about on irc this change is fine if you want to include it now or alongside your new development to make the display of validation errors much better. If you would rather include this now then +1.

jomarko:
I realized that #249 depends on this so I prefer to merge this in current state and then I will continue in work as I mentioned in BZ#1086912.

tsurdilo:
+1

PR #249, opened 2016-04-26
JBPM-5095: default gate tests

jomarko:
Depends on #248

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
@@ -487,6 +487,7 @@ public class BPMN2SyntaxChecker implements SyntaxChecker {
 			if(gateway.getGatewayDirection().getValue() != GatewayDirection.DIVERGING.getValue() && gateway.getGatewayDirection().getValue() != GatewayDirection.CONVERGING.getValue()) {
 				addError(gateway, new ValidationSyntaxError(gateway, BPMN2_TYPE, "Invalid Gateway direction for Exclusing Gateway. It should be 'Converging' or 'Diverging'."));
 			}
+			checkDefaultGate(gateway, ((ExclusiveGateway) gateway).getDefault());
 		}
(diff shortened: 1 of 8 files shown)


jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
```diff
@@ -592,6 +594,12 @@ private void checkDataObject(DataObject dataObject) {
 		}
 	}
 
+	private void checkMandatoryDefaultGate(Gateway gateway, SequenceFlow defaultSequenceFlow) {
+		if(defaultSequenceFlow == null || (gateway.getOutgoing() != null && !gateway.getOutgoing().contains(defaultSequenceFlow))) {
+			addError(gateway, new ValidationSyntaxError(gateway, BPMN2_TYPE, SyntaxCheckerErrors.NOT_VALID_DEFAULT_GATE));
+		}
+	}
+
```
tsurdilo: Default Gate should always be optional. There is no hard requirement to have it set in the runtime as well as during process modeling. It is OK if the default gate is not specified so this check method is not valid.

What you can check is if a gateway does have a default specified if indeed it is one of the outgoing flows of it. That would be OK.

jomarko: Hello @tsurdilo. I agree, I will amend some changes into #248 and then I will close this PR. Thanks.

jomarko:
Please do not merge yet.

jomarko:
Our previous conversation has gone as I amended the PR but I think now it is OK. @tsurdilo what do you think?

jomarko:
please retest this

jomarko:
jenkins retest this

tsurdilo:
@jomarko looks ok the only thing I would suggest to rename method "checkMandatoryDefaultGate" as we previously discussed the default gates are not "mandatory". checkDefaultGate would make more sense. WDYT?

jomarko:
I agree, `checkDefaulGate` is more suitable name. PR ready for review.

tsurdilo:
+1

PR #253, opened 2016-05-04
JBPM-5104: both hints using lower case

jomarko:


jlindop:
@jomarko The constants actually used in the UI are in jbpm-designer-api:
https://github.com/droolsjbpm/jbpm-designer/blob/master/jbpm-designer-api/src/main/resources/org/jbpm/designer/resources/i18n/DesignerConstants.properties,
so you need to update the string in there.

I'll email Tiho about this, as I'm not sure why we have the translations in jbpm-designer-client as well as jbpm-designer-api.

jomarko:
Thanks @jlindop, I will amend this PR soonly.

jomarko:
I think I replaced all occurrences now, please review @jlindop .

jlindop:
+1

PR #257, opened 2016-05-17
PRODMGT-1511 - Ability to copy and paste process content across windows/packages

tsurdilo:
@jlindop how to test:
1. open existing process and select part of it
2. press the "Copy to Clipboard" button in navbar (ctrl-c should work too)
3. press the "Paste" button in navbar (or ctrl-v) and it should paste as always
4. Create a new process - the paste button in navbar should be active - click it and it should paste the copied nodes)
5. Open a new tab in browser and open designer .. open existing or create new process 
6. Paste button should be enabled and you should be able to paste the copied nodes again
7. Open a new browser window and open existing or create new process
8. Paste button should be active and you should be able to paste the copied nodes again

Thanks.

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/edit.js
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/edit.js
@@ -23,6 +23,16 @@
  **/
 if (!ORYX.Plugins) 
     ORYX.Plugins = new Object();
+if (typeof(Storage) !== "undefined") {
+    Storage.prototype.setObject = function(key, value) {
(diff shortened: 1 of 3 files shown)


tsurdilo:
@jlindop just to add when copy/paste across tabs or windows the node data should also persist just like in same window paste (for example data inputs/outputs of a copied user task etc)

jlindop:
One issue: when you Paste into a BP, it doesn't seem to set the dirty flag, so if you try to save immediately after a single paste, Designer says "Process contains no changes since last save."

tsurdilo:
@jlindop nice catch. that was not working before this change either. amended

jlindop:
@tsurdilo - thanks!
+1

PR #258, opened 2016-05-18
RHBPMS_2051, 350: Fix display of i18n process and element names in si…

jlindop:
…mulation tab


tsurdilo:
+1

csadilek:
-1, sorry, test missing for fix in SimulationServlet. I do understand the desire to not test this given the convoluted logic in doPost (which also badly needs code formatting fixed) but we need to try and clean this up.

jlindop:
@csadilek  Thanks for the reminder! Unit test added - its omission was just an oversight. We're looking into reformatting the code as there are other files in designer which also need it.

csadilek:
+1

PR #264, opened 2016-05-26
RHBPMS-596 - Message and Service task default properties displayed when re-open process

tsurdilo:


jlindop:
Tested and works fine.
+1

PR #265, opened 2016-05-26
RHBPMS-596 - Message and Service task default properties displayed when re-open process

tsurdilo:


jbpm-designer-backend/src/test/java/org/jbpm/designer/test/bpmn2/Bpmn2UnmarshallingTestCase.java
```diff
@@ -48,17 +49,18 @@ private static File getTestJsonFile(String filename) {
...
+//    public void testSimpleDefinitionsUnmarshalling() throws Exception {
+//        Bpmn2JsonUnmarshaller unmarshaller = new Bpmn2JsonUnmarshaller();
+//        Definitions definitions = ((Definitions) unmarshaller.unmarshall(getTestJsonFile("empty.json"), "").getContents().get(0));
+//        assertEquals("<![CDATA[my doc]]>", definitions.getDocumentation().iterator().next().getText());
+//        assertEquals("http://www.w3.org/1999/XPath", definitions.getExpressionLanguage());
+//        assertEquals("http://www.omg.org/bpmn20", definitions.getTargetNamespace());
+//        assertEquals("http://www.w3.org/2001/XMLSchema", definitions.getTypeLanguage());
+//        assertTrue(definitions.getRootElements().size() == 1);
+//        definitions.eResource().save(System.out, Collections.emptyMap());
+//    }
```
hasys: I guess we have to remove tests for not implemented functionality for 6.4.X branch. In my opinion there are no reasons to store it here in this way.

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
```diff
@@ -4304,7 +4305,7 @@ protected void applyStartEventProperties(StartEvent se, Map<String, String> prop
     }
 
     protected void applyMessageProperties(Message msg, Map<String, String> properties) {
-        if(properties.get("name") != null) {
+        if(properties.get("name") != null && properties.get("name").length() > 0) {
```
hasys: Maybe predicate function will be good for this purpose something like:
`private boolean isPropertyEmpty(Map<String, String> properties, String propertyName) {
    return properties.get(propertyName) != null && properties.get(propertyName).length() > 0;
}` 
If we will use it we never will meet issues like this again. And if something will be changed we can easily modify it in one place.

jlindop:
Tested and works fine.

jbpm-designer-backend/src/test/java/org/jbpm/designer/test/bpmn2/Bpmn2UnmarshallingTestCase.java
tsurdilo: Sure I will remove them instead of commenting. We should align the testing code from master and 6.4.x soon.

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
tsurdilo: Agreed but not in this PR since this is 6.4.x. Let's get this change trickle down to 6.4.x through master via a separate PR specific to this refactoring.

tsurdilo:
@jlindop amended (removed disabled tests per @hasys recommendations)

jlindop:
+1

csadilek:
Isn't this also removing tests that previously did run (that weren't disabled) i.e. testSimpleDefinitionsUnmarshalling?

tsurdilo:
@csadilek i removed tests that don't make sense any more to run. Our master branch has alot more tests and I will find commits to cherry pick to 6.4.x so that they are the same. The tests I removed are either not in master or have already been updated there.

csadilek:
@tsurdilo the following tests were active on 6.4.x and are currently on master [1] but have been removed in this commit on 6.4.x:

testSimpleDefinitionsUnmarshalling
testPoolUnmarshalling
testWorkItemHandlerNoParams
testDocumentationForSwimlane

[1] https://github.com/droolsjbpm/jbpm-designer/blob/master/jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2UnmarshallingTest.java

tsurdilo:
@csadilek 
testWorkItemHandlerNoParams -- revised and added back
testDocumentationForSwimlane - revised and added back
testMessageUnmarshalling - renamed to testNoDefaultMessageCreated and revised
testSimpleDefinitionsUnmarshalling - revised and added back

testPoolUnmarshalling - we dont use pools so is ok to remove . i will do the same in master asap

other removed tests no longer apply

PR #276, opened 2016-06-01
JBPM-5135: added tests for service tasks marshalling and syntax checking

jomarko:


jlindop:
Looks good - thanks for adding more tests!
+1

PR #281, opened 2016-06-02
JBPM-5143 - MI Subprocess mi data input and mi data output properties should be String type

tsurdilo:
@jlindop to test just drag/drop a MI subprocess onto canvas and make sure the MI data input and MI data output properties are String type (not dropdowns any more).


jlindop:
+1

PR #286, opened 2016-06-08
RHBPMS-678: parameters of conditions can not be null more

jomarko:


jlindop:
+1

jomarko:
@jlindop Very good point. Thanks for notice. Can you please review again? String INVALID_CONDITION_ERROR should not be displayed more.

jlindop:
Looks good.
+1

PR #287, opened 2016-06-08
(6.4.x) RHBPMS-678: parameters of conditions can not be null more

jomarko:


jlindop:
I don't know whether we want this in the 6.4.x branch - RHBPMS-678 is a regression for v6.2.0, but it's not a blocker. What does one of the gatekeepers think?

jomarko:
I understand, I am fine if the PR will be not merged into 6.4.x it is not critical bug.

jomarko:
In case we want to include this PR into 6.4.x I amended this PR according to comments on master.

jomarko:
PR created for 6.5.x.

PR #290, opened 2016-06-10
RHBPMS-4011: Fix i18n Simulation problem in IE11

jlindop:
Note: RHBPMS-4011 has the flags for release in 6.3.1.

To test, I do a mvn -Dfull build of jbpm-designer, and a mvn -Dfull build of kie-wb-distributions/kie-wb (which takes a while), then run the kie-wb-6.4.1-SNAPSHOT-eap6_4.war in EAP and test from IE11 on Windows 7.


hasys:
+1. Looks good for IE11 and FF38 on Windows 8.

PR #291, opened 2016-06-10
JBPM-5153 SimulationServletTest.testLocalizedStartEndTime fails on some JDKs

baldimir:
- Fixed testLocalizedStartEndTime test.

I tried changing the language, but didn't find any that match the needs of this test (non latin, not changing months on different JDKs)


jlindop:
Thanks for this fix.
+1

PR #293, opened 2016-06-13
latest Zanata translation changes in jbpm-designer from 2016-06-13

kiereleaseuser:


jlindop:
Although it looks as though some translations of ORYX.I18N.Grouping.ungroupDesc have been removed, these translations are marked as 'fuzzy' in Zanata, and so will be added back to the properties files after they have been approved by translators.
+1

PR #298, opened 2016-06-18
JBPM-5159 - Japanese text get garbled in PDF generated from Business Process

tsurdilo:
this also removes the org.apache.xmlgraphics.fop depends from pdf generation (using itext now everywhere)


psiroky:
We should also remove the FOP from kie-parent (https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/pom.xml#L87 + depMgmt). I believe designer was the only place where it was used.

tsurdilo:
@psiroky you are right - will do a pr for that as well. thanks!

tsurdilo:
@psiroky https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/224

jlindop:
@tsurdilo  This fix works, but the text in PDF's is a little blurry. I wonder whether you can set some parameters or hints to improve it.

tsurdilo:
@jlindop its png image inside the generated pdf. I resize it to fit the A4 page size of the pdf doc. I think what you see affects mostly small test processes but for larger ones it should look fine. If it's readable I would go with it for now and improve later/soon. WDYT?

jlindop:
@tsurdilo  This seems reasonable. I'll create a separate Jira for the issue of how sharp the PDF is.
+1

tsurdilo:
@jlindop ok sounds good - here is example of evaluation process it seems ok to me but yeah the resizing of the image to fit the page size could be improved in your new jira you will create. thanks! 

-- http://imgur.com/BnJyLPn

PR #303, opened 2016-06-23
RHBPMS-4021 - Not able to open BP without browser refresh

tsurdilo:
@jlindop to test you can follow this video: http://screencast.com/t/bv8luutHA1V
@jlindop I have also tested to make sure this does not affect https://bugzilla.redhat.com/show_bug.cgi?id=1269969 but since you made that change could you also test that one again? Thanks.

### DIFF
--- a/jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerPresenter.java
+++ b/jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerPresenter.java
@@ -179,6 +179,11 @@ public class DesignerPresenter
     }
 
     public void onLockChange( @Observes UpdatedLockStatusEvent updateLockEvent ) {
+        if(updateLockEvent.isLocked()) {
+            view.setViewLocked(true);
+        } else {
+            view.setViewLocked(false);
+        }
         view.raiseEventUpdateLock();
     }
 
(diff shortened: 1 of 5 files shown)


csadilek:
@tsurdilo this change seems wrong to me. The parameter needs to be true (as it was before) and hence trigger a reload. If you open the designer and the asset is locked it will be in read-only mode. If the lock goes away (the other user is done editing) a refresh needs to happen to put the designer back in edit mode. This is also the original behaviour (before @jlindop's change).

csadilek:
-1 based on comment above until clarification. Just a reminder that when testing asset locking 2 different users in 2 different browsers need to be used.

jlindop:
@tsurdilo  This change does stop the asset locking working - if two users have a BP open and one of them edits the BP, when this user closes the BP without saving, the other user's editor remains readonly, whereas it should re-open in edit mode. 
assetlock

jlindop:
@tsurdilo When process C is opened, ORYX.READONLY is true at https://github.com/droolsjbpm/jbpm-designer/blob/6.4.x/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Core/main.js#L199
I suspect this is the cause of the problem.

Following the reproduction steps for RHBPMS-4021, if you close process B before opening C, there's no problem. This is because in this case ORYX.READONLY is false at the line above, as you'd expect.

PR #304, opened 2016-06-23
RHBPMS-4021 - Not able to open BP without browser refresh

tsurdilo:
@jlindop to test you can follow this video: http://screencast.com/t/bv8luutHA1V
@jlindop I have also tested to make sure this does not affect https://bugzilla.redhat.com/show_bug.cgi?id=1269969 but since you made that change could you also test that one again? Thanks.

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/saveplugin.js
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/saveplugin.js
@@ -177,7 +177,7 @@ ORYX.Plugins.SavePlugin = Clazz.extend({
             if (this.editorLocked && !parent.isLockedByCurrentUser()) {
                 this.editorLocked = false;
             } else if (!this.editorLocked && !parent.isLocked()) {
-                ORYX.EDITOR.updateViewLockState(true);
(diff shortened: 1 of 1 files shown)


csadilek:
@tsurdilo this change seems wrong to me. The parameter needs to be true (as it was before) and hence trigger a reload. If you open the designer and the asset is locked it will be in read-only mode. If the lock goes away (the other user is done editing) a refresh needs to happen to put the designer back in edit mode. This is also the original behaviour (before @jlindop's change).

csadilek:
-1 based on comment above until clarification. Just a reminder that when testing asset locking 2 different users in 2 different browsers need to be used.

tsurdilo:
@csadilek the parameter set to true here triggers an reload loop in the condition described in the jira. From testing with multiple users for me it did not affect the current behavior and fixed this particular issues. When testing with other editor types (drl for example) this infinite reload loop does not happen so if the solution here is to you wrong could you please tell me how that is handled there so we can mimic it in designer as well? Thanks.

csadilek:
@tsurdilo please see @jlindop's description on your other pr which confirms my comment above: https://github.com/droolsjbpm/jbpm-designer/pull/303

PR #311, opened 2016-06-29
RHBPMS-1698 - Build & Deploy fails due to duplicate ProcessID after copying business process 

tsurdilo:


jlindop:
Tested and works fine.
+1

PR #313, opened 2016-06-30
JBPM-5162 - default gate is not readable just after selecting a sequence flow which has Japanese name

tsurdilo:


jlindop:
@tsurdilo   If you select a default gate, click away from the gateway, then click on the gateway again, the 'Default gate' property shows just the ID, where it used to show "Name:ID".

PR #317, opened 2016-07-01
RHBPMS-4034: Show default gateway name:id in combo after reselection …

jlindop:
…in properties view

@tsurdilo  This makes sure the default gateway "name : id" is shown in the combo after the user selects a default gateway, clicks away from the gateway, then selects the gateway again.


tsurdilo:
+1

PR #319, opened 2016-07-06
RHBPMS-511: Add default Name to UserTasks "Task_n" when editing BP.

jlindop:


jlindop:
@tsurdilo   I'm not sure why the commit for 'shape.js' isn't showing the diff. If you look at the file, searching for "updateLastUserTaskID" shows you the changes.

I also looked into whether to add a default name in the marshaller, but because of the way I've done it in the editor, there shouldn't be any UserTasks without a name.

jlindop:
Closing after discussions with Tiho.

PR #321, opened 2016-07-06
JBPM-5192 - After the simulation fails the server becomes unresponsiv…

tsurdilo:
…e and it does not even respond to shutdown requests


mswiderski:
+1 to merge

PR #340, opened 2016-07-07
RHBPMS-4034: Show default gateway name:id in combo after reselection …

jlindop:
…in properties view (#317)
Commit cherry-picked from 6.4.x


tsurdilo:
+1

PR #345, opened 2016-07-11
JBPM-5179: i18n - Fix generation of task form via "Edit Task Form" bu…

jlindop:
…tton for task with non-Latin chars
cherry-picked from 6.5.x


csadilek:
+1

PR #353, opened 2016-07-14
JBPM-4996: shown No defined type in process documentation

jomarko:


jlindop:
+1

manstis:
Is this completely untestable? IIRC Designer has some special dispensation from Unit Tests for JavaScript changes....?? @csadilek please correct me if I'm wrong...?

csadilek:
@manstis yes, designer got an exception for .js changes. I am not happy with that, as it's definitely not untestable, just a matter of putting in the effort, but that wasn't my call.

PR #354, opened 2016-07-15
JBPM-5210 - Link each element in Process Documentation to the actual element in the Process Model

tsurdilo:
@jlindop JS changes only . Adds a "View in Model" button for each element of the documentation. When clicked switches the tab to Process Model and selects that particular element for user.

see http://imgur.com/rWSHHXq


jlindop:
+1

PR #357, opened 2016-07-15
JBPM-4988: connectors with name, documentation or condition added int…

jomarko:
…o process documentation

### DIFF
--- a/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/view.js
+++ b/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/view.js
@@ -2201,14 +2201,14 @@ ORYX.Plugins.View = {
     },
 
     findShapeWithId : function(shape, nodeid) {
-        if(shape instanceof ORYX.Core.Node) {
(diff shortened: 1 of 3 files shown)


jomarko:
@tsurdilo @jlindop hello to all, this PR adds into process documentation connectors which have set 'name', 'documentation' or 'condition expression'. Problem of this PR is that if there is no such connector, there will be empty section 'Connectors' in section 'Element Details'. Do you have any idea how to enhance my PR? Thanks for any ideas.

tsurdilo:
@jomarko I will look at this asap. also if we add connectors there needs to be a change to 
https://github.com/droolsjbpm/jbpm-designer/commit/1f23e8529ebb05ac9216008902e12478ba84a09e 
to also include Edges that i need to change too

jomarko:
Hello @tsurdilo , I updated the method `findShapeWithId` to assume also ORYX Edges in addition to Nodes. Let me know what do you think.

https://github.com/jomarko/jbpm-designer/blob/4fd04925b7b88a1353cb5df1e80c4cbb4c5491bb/jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/view.js#L2203-L2216

tsurdilo:
@jomarko looks good. thanks
+1

PR #368, opened 2016-07-22
[master] JBPM-5217 - Link each element of the process model to its location in  the process documentation

tsurdilo:
(JS changes only)

@jlindop : adds a new button in the shape menu for each node in the process.
when the new documentation icon in shape menu is clicked, the tab changes from process modeling to process documentation and the page scrolls the location where this element is defined.

Note: for sequence flows we only show sequence flows in the docs that have either their name, documentation or conditionexpression properties set. If the shape menu docs button is clicked on a sequence flow which does not have at least one of these elements set, designer shows a notification telling the user that element does not exist in the docs and does not switch to the docs tab. if the sequence flow has at least one of these elements the same applies as for any other node type.


tsurdilo:
@jlindop amended per your code comment

jlindop:
+1

PR #369, opened 2016-07-22
[6.5.x] JBPM-5217 - Link each element of the process model to its location in the process documentation

tsurdilo:
(JS changes only)

@jlindop : adds a new button in the shape menu for each node in the process.
when the new documentation icon in shape menu is clicked, the tab changes from process modeling to process documentation and the page scrolls the location where this element is defined.

Note: for sequence flows we only show sequence flows in the docs that have either their name, documentation or conditionexpression properties set. If the shape menu docs button is clicked on a sequence flow which does not have at least one of these elements set, designer shows a notification telling the user that element does not exist in the docs and does not switch to the docs tab. if the sequence flow has at least one of these elements the same applies as for any other node type.


jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/shapemenu.js
```diff
@@ -311,6 +324,28 @@ ORYX.Plugins.ShapeMenuPlugin = {
...
+			}
+		}
+		if(showInDocs) {
+			ORYX.PROCESSDOC_RESOURCEID = this.currentShapes[0].resourceId;
+			Ext.getCmp('maintabs').setActiveTab(2);
+		} else {
+			this.facade.raiseEvent({
+				type 		: ORYX.CONFIG.EVENT_NOTIFICATION_SHOW,
+				ntype		: 'info',
+				msg         : ORYX.I18N.ShapeMenuPlugin.unableToFindNodeSource,
```
jlindop: Should this be  ORYX.I18N.ShapeMenuPlugin.unableToLinkElementToDocs ?
For this constant, maybe "Element not included in documentation" might be better, as the current message "Unable to..." makes it sound as though there's an error.

tsurdilo: Yeah that's better wording. Will change

tsurdilo:
@jlindop amended per your code comment

jlindop:
+1

PR #373, opened 2016-07-25
[6.4.x] RHBPMS-4164 - org.jbpm.designer.server.BusinessProcessCopyHelperTest fails on Windows

tsurdilo:
updated test

### DIFF
--- a/jbpm-designer-backend/src/test/java/org/jbpm/designer/server/BusinessProcessCopyHelperTest.java
+++ b/jbpm-designer-backend/src/test/java/org/jbpm/designer/server/BusinessProcessCopyHelperTest.java
@@ -34,6 +34,7 @@ import org.uberfire.java.nio.base.options.CommentedOption;
 import static org.junit.Assert.*;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.*;
+import static org.hamcrest.CoreMatchers.containsString;
 
 @RunWith(MockitoJUnitRunner.class)
 public class BusinessProcessCopyHelperTest {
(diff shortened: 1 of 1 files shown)


winklerm:
Hi @tsurdilo, thanks for fixing the test! However, I was wondering why the process ID is different on Windows than on Linux, is it caused just by the test (i.e. not by {{BusinessProcessCopyHelper}})?

From my point of view, I would expect the process ID to be the same on all platforms and the test should test also that to prevent future regressions. Have I overlooked something?

Thanks for considering it!

tsurdilo:
@winklerm I think you make a good point but also think that we need to look at this difference further up. BusinessProcessCopyHelper receives 2 paths (source and destination) and applies the same code on this path regardless of the OS.  I think is better to leave this as is for now and for me to look into the workbench code to look for clues why paths are different. wdyt?

csadilek:
@tsurdilo @jlindop is this ready to go? Don't see a plus one.

tsurdilo:
@csadilek I think its good to go, just maybe waiting on @winklerm to reply if he has a minute

winklerm:
@tsurdilo ok, I have created RHBPMS-4175 for the investigation. It seems strange to me since I was not able to replicate the different process ID manually in business-central, so it might be just a test problem.

On the other hand, looking at [1], I wonder if Path.toString() method returns '/' or '\' on Windows. That would mean that the _BusinessProcessCopyHelper_ is the culprit.

[1] https://github.com/droolsjbpm/jbpm-designer/commit/ab8efc5a44f419390b4f3a26017946fefb76841a#diff-8d31dc2f6e2c0d8b4aa3a58f3bc33ca4R89

@tsurdilo @csadilek +1 to merge.

winklerm:
I just realized that the _Path_ in the code refers to _org.uberfire.backend.vfs.Path_, not _java.nio.Path_. So my previous comment might be completely off-topic, sorry :)

PR #377, opened 2016-08-05
BPMSPL-295 - Higher level of automation when importing custom workItemHandlers from a service repository

tsurdilo:


tsurdilo:
note this also fixed the test failures on master  (BPMN2SyntaxCheckerTest specifically)

jomarko:
Hello @tsurdilo , thanks for your PR. I fetched your PR locally and I have found few issues. Before this PR will be merged can you please take a look at the issues?

1. Whenever I save file, I obtain two messages about successful saving of the file. If you confirm that this problem comes from master branch and not from your PR, I will file separate JIRA for this issue.
screenshot from 2016-08-08 13-49-23

2. Service repository editor doesn't load icons from the service repository: file:///<pathToDesignerRepo>/jbpm-designer-backend/src/test/resources/org/jbpm/designer/web/server/servicerepo
screenshot from 2016-08-08 13-56-53

tsurdilo:
@jomarko thanks for the great feedback!
here are some comments:

1) these error message are different. One is for the save of the business process the other is specific for the business process "image" (SVG) (which is stated in the message too. We enabled the saving of the process SVG when a process is also saved in  https://github.com/droolsjbpm/jbpm-designer/commit/4c722154bca157a4052a3a826762bff84f22dc2e (this was false by default before so the second message was not seen unless users manually enabled id)

tsurdilo:
@jomarko 
i can reproduce 2) as well. fixing and amending now

tsurdilo:
@jomarko 

amended per your request for 2)

icons from file system should work now. also added that if there is no icon property defined in wid or if its an invalid location (icon cannot be found)  the default service task icon will be displayed in its place (this is consistent as when the service node is installed designer will assign it the default service icon if the user defined icon in wid cannot be located)

psiroky:
Minor thing - we should _not_ be using the BPMSPL Jira project in commits

tsurdilo:
@psiroky thanks - looking at that jira im not sure what other value to use. let me know.

tsurdilo:
@jomarko @jlindop if you have time please check when you can so we can get this in.

psiroky:
@tsurdilo I think we agreed to create upstream JIRA (jBPM in this case) for every BPMSPL and we can reference that.

tsurdilo:
@psiroky ok i can create one then. was hoping there would be one already but only reference is the customer support case for this.

jomarko:
@tsurdilo I am looking right now.

tsurdilo:
@psiroky created JBPM-5249 and amended commit message. thanks

tsurdilo:
@psiroky updated commit message for https://github.com/droolsjbpm/jbpm-console-ng/pull/451 as well

psiroky:
@tsurdilo thanks!

tsurdilo:
test fails not related to this PR

(org.jbpm.designer.web.server.SimulationServletTest.testRunSimulationLocalizedNames
org.jbpm.designer.web.server.SimulationServletTest.testLocalizedStartEndTime)

tsurdilo:
jenkins retest this

jomarko:
So here are my comments.

1.  Because of this bug: https://issues.jboss.org/browse/JBPM-5229, I am not able to verify the most important use case. It means, I am able to import WID and check if pom.xml and DD were changed, but I am not able to start the modeled process, that is using imported custom task.

2. In the method [1] is parsed list of task names. There is parsed also some category but my tries to use it were not successful. Can you please provide more info how this should work, because this is not mentioned in appropriate planing jira.

3. In the planing jira you mentioned, that the variable org.jbpm.service.repository will cause default value also for service repository, but currently it doesn't hold true

4.In the planing jira is mentioned requirement to have possibility specify as default handler so parameters for constructor of this handler. Currently, if I open DD editor and there was imported WID file with attribute "parameters", these parameters don't seem to be used in the DD. Maybe I just do not understand correctly attribute "parameters" in the WID file and how to instantiate handler(that have constructor with parameters) in DD.

5. Currently if "defaultHandler" attribute doesn't contain resolver type, no resolver type is also added into DD. However, in the planing jira I saw requirement to assume default handlers without resolver type as handlers with MVEL resolver type. So, is current behavior correct?

[1]
https://github.com/droolsjbpm/jbpm-designer/pull/377/commits/5d977a0e0d9cb192ed449484a6f38c36d57816ba#diff-47eb79f39bcd0c584a5f7602d0052eb7R525

tsurdilo:
@jomarko 

1. @krisv is there any workaround for this currently so @jomarko could finish his test for this?
2. This was looking ahead a little where users could specify a task name and its category for lookup. So for example:

3. The wording might have been somewhat wrong (I will make this clear in the docs) . The value is added to the dropdown list:

screen shot 2016-08-09 at 9 59 02 am


5. It will be handled same way as before where  no resolvertype option was even available.

tsurdilo:
@jomarko 
For 1 (https://issues.jboss.org/browse/JBPM-5229) - this is NOT a bug. You need to have an execution server set up before you test (workbench is for authoring only now). So without that set up on your end you will not to complete your tests (but this is not a real bug) -- see more here http://mswiderski.blogspot.com/2016/07/jbpm-v7-workbench-and-kie-server.html


@jomarko 
for 4 - The "parameters" is not the input parameters to the task but for the handler. For MVEL type resolver all of the parameters can already be specified inside the defaulthandler String, for example:

"defaultHandler" : "new com.something.MyHandler(a,b,c,d)"

and this will be automatically resolved. For REFLECTION type handler the parameters are way too hard to be resolved programatically and should only be entered through the DD editor. But for users it makes more sense to just use MVEL resolver and enter parameters into a single string anyways.

tsurdilo:
@jomarko I answered all your comments and i think there are no bugs with anything. Please let me know if need any help setting up execution server so you can complete that part of your test

csadilek:
@tsurdilo there are two test failures (SimulationServletTest). Can you look into those, please.

tsurdilo:
@csadilek yes I am looking at that but it's not related to this PR. 

screen shot 2016-08-09 at 10 52 44 am

jomarko:
@tsurdilo thanks for clarification, I agree with merge to speed up cherry pick for 6.5.x.

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/server/JbpmServiceRepositoryServlet.java
```diff
@@ -227,22 +248,56 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp)
...
+
+	public boolean canUpdateConfigForWorkitem(WorkDefinitionImpl workitem) {
+		return workitem != null &&
+				workitem.getName() != null &&
+				workitem.getName().trim().length() > 0 &&
+				workitem.getDefaultHandler() != null &&
+				workitem.getDefaultHandler().trim().length() > 0;
+	}
+
+	protected static byte[] getBytesFromFile(File file) throws IOException {
```
csadilek: Why not use org.apache.commons.io.IOUtils.toByteArray or java.nio.file.Files.readAllBytes...

csadilek: There should be no reason for standard functionality like this to be in the designer esp. as we can use existing libraries already on the classpath.

csadilek:
One comment above. Also, the test coverage seems pretty low i.e. for ServiceRepoUtils. Can this be increased?

tsurdilo:
@csadilek sure thing. working on it

jomarko:
Hello @tsurdilo, I checked again importing service tasks on startup in the form:
-Dorg.jbpm.service.servicetasknames=MyTask:MyCategory1,MyOtherTask:MyCategory2

Unfortunately this causes that no custom task is installed into workbench. I suggest to completely remove code, that assumes possibility to specify also categories for service tasks. I think there are two reasons for that.

1. it is not mentioned in the requirement
2. the requirement is quite complex and this will help to keep the PR at the manageable size.

Importing in the form below works perfectly.
-Dorg.jbpm.service.servicetasknames=MyTask,MyOtherTask

What do you think?

tsurdilo:
@jomarko yep agreed. have removed it and still adding tests. will amend soon

tsurdilo:
@csadilek 
Added ServiceRepoUtilsTest and added test to JbpmServiceRepositoryServletTest

@jomarko the requested code has been removed

let me know if there is anything else guys haha :)

tsurdilo:
@jomarko Can you please +1 this ? thanks

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/server/JbpmServiceRepositoryServlet.java
```diff
@@ -227,22 +250,30 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp)
...
 		}
 	}
 
-    private String getRepositoryDir(String uuid) {
-        int iStart = uuid.indexOf("//");
-        iStart = uuid.indexOf('/', iStart + 2);
-        int iEnd = uuid.lastIndexOf('/');
-        return uuid.substring(iStart, iEnd);
-    }
+	public boolean canUpdateConfigForWorkitem(WorkDefinitionImpl workitem) {
```
jomarko: Please remove this method. Its copy is present and used in ServiceRepoUtils class.

jomarko:
Sorry for being late with review. Our working places are too far away.

However, thanks for removing the mentioned code and some nice tests. I will add some additional tests as one part of the requirement verification.

Here is my ack: +1
Please, just remove the method from the comment above at first.

tsurdilo:
@jomarko done.

tsurdilo:
@csadilek ready to be merged.

csadilek:
great team work @tsurdilo @jomarko, thanks ;).

PR #388, opened 2016-08-19
JBPM-5272: WorkDefinitions.wid installed also into not empty locations

jomarko:


tsurdilo:
@jomarko 
very sorry but -1  -- we do not want to install the default wid in this case because it may overwrite workitems already defined by user (as later all wids entries are merged to be displayed on the shape menu). This works as intended.

PR #392, opened 2016-08-23
[6.5.x] JBPM-5278: added tests and refactoring done

jomarko:
This PR adds tests for automatic WID import feature.

Next 3 main changes of this PR are:

1. Injection of services needed to change pom.xml and DD was moved to lower level in the architecture. As consequence of this some interfaces are now more simple.

2. There is added try-catch block in the JbpmPreprocessingUnit class. This try-catch block ensures that installation of WIDs will continue even if the installation of some WID failed.

3. There is replaced method `deleteAsset` with the method `deleteAssetFromPath`. The method `deleteAsset` was not working and was returning `false` as information about not successful deletion. As consequence of the method replacement, the user will not more see the message, that the WorkItem is already installed. Instead of this the old *.wid file will be overwritten withe new one. Also the icon will be overwritten if the new wid references the same icon.

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
@@ -708,16 +708,20 @@ public class JbpmPreprocessingUnit implements IDiagramPreprocessingUnit {
                 List<String> toInstallTasks = Arrays.asList(defaultServiceRepoTasks.split("\\s*,\\s*"));
                 for(String installTask : toInstallTasks) {
                     if(workitemsFromRepo.containsKey(installTask)) {
-                        ServiceRepoUtils.installWorkItem(workitemsFromRepo,
-                                installTask,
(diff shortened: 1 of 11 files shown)


tsurdilo:
@jomarko I will test this asap and would consider this for master only and because of the changes not for 6.5.x because the preprocessing unit changes are major and could introduce some side effects that the tests do not cover. Please do not hold https://issues.jboss.org/browse/BPMSPL-295 because of this as I do think its good for master atm and would go in into the next release anyways.

tsurdilo:
@jomarko I don't see pr for this for master branch, am i wrong? if so please add and i will test asap

jomarko:
@tsurdilo thanks for your comments. I created PR for master branch. But I would ask you to keep this PR open. I will omit refactoring for 6.5.x branch. There will remain only tests which are necessary for BPMSPL-295. Thanks.

jomarko:
@tsurdilo the majority of changes is now only in tests. Please take a look when you have a moment. Thanks.

tsurdilo:
+1

jomarko:
Conflicts resolved, jenkins please retest this.

PR #395, opened 2016-08-27
RHBPMS-4196 - It is not possible to open copy of Business Process if …

tsurdilo:
…it's name contain whitespaces or multybyte chars


krisv:
+1

PR #396, opened 2016-08-27
[6.5.x] RHBPMS-4196 - It is not possible to open copy of Business Process if …

tsurdilo:
…it's name contain whitespaces or multybyte chars


hasys:
Hi @tsurdilo,
works good, but there are some inconsistency now. If you create new file with localized name it will have ID like on a top of attachment, if you make a copy with localized name it will be nicer (bottom part), but in my opinion it will be better if both ways will produce the same ID. WDYT is it possible to reuse existing code from AssetServiceServlet for working with files, or reuse this code in both places? For example some Util class with unified ID generation.
untitled

tsurdilo:
@hasys amended  - we have to push the process id on copy through the same code as on new process creation so in both cases it will generate an id similar to your first picture above (the non-readable one).

krisv:
+1

PR #403, opened 2016-08-31
JBPM-5300 - Enable custom assignments for service tasks only when cus…

tsurdilo:
…tomeditors property is not empty (if it exists)

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
@@ -402,7 +402,7 @@ public class JbpmPreprocessingUnit implements IDiagramPreprocessingUnit {
     }
 
     @SuppressWarnings({ "unchecked", "rawtypes" })
-    private void evaluateWorkDefinitions(Map<String, WorkDefinitionImpl> workDefinitions, Asset<String> widAsset, String assetLocation, Repository repository, IDiagramProfile profile) throws Exception {
+    public void evaluateWorkDefinitions(Map<String, WorkDefinitionImpl> workDefinitions, Asset<String> widAsset, String assetLocation, Repository repository, IDiagramProfile profile) throws Exception {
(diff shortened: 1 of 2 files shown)


tsurdilo:
jenkins build fails seem unrelated to designer completely

jomarko:
@tsurdilo seems to be good. Just a question, what caused the decrease of globalAssets and defaultStuff?

tsurdilo:
@jomarko not sure I understand your question. if you ask about why in case customeditor is defined in wid properties are shown as individual inputs/outputs rather than io/editor? 

We have customers that do not wish to use the data i/o editor for custom service tasks. rather they would like to plug in their own or make it simpler by just having each data input / output as individual properties on the panel -- then their users don't have to worry about knowing whats input whats output etc etc. Also then they can plug in their own really custom editor for each property as well if they would like as designer allows for fully custom editors for properties. 

HTH

jomarko:
@tsurdilo thanks for comment. Sorry for my unclear question. Here it is again:

I was curious, why there are following two changes:
`assertEquals(31, globalAssets.size());` - `assertEquals(30, globalAssets.size());`
`assertEquals(3, defaultStuff.size());` - `assertEquals(2, defaultStuff.size());`

tsurdilo:
@jomarko due to update to extend RepositoryBaseTest

hasys:
Hi @tsurdilo,
> When customEditor is defined in wid, designer does not show data assignments as do other tasks ,but parses the wid and creates individual property for each input/output and that marshallers/unmarshallers also know how to deal w/ that situation.
HTH
It looks like individual properties shows only for input variables ('parameters' definitions in WID file). Individual properties for output variables ('results' definitions in WID file) is not appear in Task properties bar. Is it an expected behavior?

tsurdilo:
@hasys yes iirc that was the expected behaviour with custom editors you deal with inputs only

tsurdilo:
jenkins retest this

jomarko:
Fail seems to be unrelated to this PR. The fix is exactly the same as the proposed in the email communication between me, @tsurdilo and @hasys. +1

csadilek:
jenkins retest this.

PR #415, opened 2016-09-06
JBPM-5300 - Enable custom assignments for service tasks only when cus…

jomarko:
…tomeditors property is not empty (if it exists)

This PR is cherry-pick of #403. QE team would like to include this change also into 6.5.x branch because our effort is to minimize work with updating of internal tests.


tsurdilo:
+1

PR #418, opened 2016-09-06
Decrease the number of GWT workers

psiroky:
* the project should sucessfullly compile
   on a machine with 8GB RAM (even that can be
   seen as quite a lot). Running 4 parallel
   workers on 8GB resulted in OOMEs.


csadilek:
+1

PR #420, opened 2016-09-06
RHBPMS-1560: i18n the BP save dialog

jlindop:
Note the save function has been restructured because 'Ext.MessageBox.prompt' is non-modal, whereas the previously used function 'prompt' was modal.


jlindop:
Note that the diff shown for this PR makes it look as though more has changed than it has. I've refactored the 'save' function to allow for Ext.MessageBox.prompt being non-modal, but haven't changed what saving actually does.

tsurdilo:
+1

PR #423, opened 2016-09-06
[6.4.x] GUVNOR-2627 - Process Designer: Remove Cancel End Event from pallete

tsurdilo:


krisv:
Afaik this should only go in master and 6.5.x, not aware of a 6.3.x patch update request for this.

PR #431, opened 2016-09-07
RHBPMS-4201: Show simulation path images in Chrome.

jlindop:


jbpm-designer-client/src/main/resources/org/jbpm/designer/public/simulation/pathimage.jsp
```diff
@@ -8,6 +8,18 @@
 <script>
     function showPathImage() {
         var pathSVG = parent.getPathSVG();
+        var isChrome = /Chrome/.test(navigator.userAgent) && /Google Inc/.test(navigator.vendor);
```
hasys: Is it possible to add other chromium based browsers at least for community version?

jlindop:
@hasys  There are other places in designer with fixes for Chrome, so if we want to extend coverage to other browsers, then I think it would make sense to do this in a separate Jira.

tsurdilo:
+1

PR #434, opened 2016-09-07
[6.5.x] RHBPMS-1566 - Scripts with || are incorrectly saved

tsurdilo:


jomarko:
I think this piece of code is worth to extract into separate method. I found 9 occurrences in the Unmarshaller.


                String scriptLanguage;
                if(properties.get("script_language").equals("java")) {
                    scriptLanguage = "http://www.java.com/java";
                } else if(properties.get("script_language").equals("mvel")) {
                    scriptLanguage = "http://www.mvel.org/2.0";
                } else if(properties.get("script_language").equals("javascript")) {
                    scriptLanguage = "http://www.javascript.com/javascript";
                } else {
                    // default to java
                    scriptLanguage = "http://www.java.com/java";
                }

tsurdilo:
@jomarko agreed. would it be ok to make that part of a refactoring PR separate from this? I can get that in asap.

jomarko:
@tsurdilo I am fine with refactoring in separate PR, if the new PR will me applied as for master so for 6.5.x. Refacotring only one branch could cause problems during cherry-picking between these two branches. So, do you plan to refactor both branches?

jlindop:
Tested and works well, so ready to approve if the refactoring discussed above is done separately.

tsurdilo:
@jomarko I will create a jira for it and it will be included whenever a release asks for it. I understand the value of refactoring this code but 
* it is not part of the fix for this platform jira
* it has been around for a number of years and does not right now for this jira add particular value for it
* it will be in master as soon as we get our other priority jiras in place for the next release

HTH

jomarko:
+1 from myself, probably also from @jlindop according to his previous comment.

csadilek:
I don't understand why this can't be refactored now. It looks like a matter of minutes to me. We certainly shouldn't be postponing the refactoring to (maybe) the next release. There's always gonna be something that will look to be more important than the refactoring. So it will never get done. We have empirical evidence for that :).

jomarko:
@csadilek @tsurdilo I am ready and willing to do the mentioned refactoring. Just let me know if I have green to do it as on master so for 6.5.x. I understand that @tsurdilo have a lot of work with other Jiras and I would like to help.

csadilek:
@jomarko definitely a green light from my side ;). thanks.

tsurdilo:
@jomarko same here :) was never a question if we do this or not, just in what pr and when. Thanks for your help!

PR #442, opened 2016-09-09
JBPM-5315 RHBPMS-4251 Removed usage of IOUtils.closeQuietly when closing Scanner instance.

baldimir:
- Substituted use of IOUtils.closeQuietly with scanner.close() so it works on JDK 1.6. 

No master PR, because jdk 1.6 currently not relevant there.


tsurdilo:
+1

PR #447, opened 2016-09-09
RHBPMS-519: Fix problem selecting gateway in BP

jlindop:
Fixed by making the call from propertywindow.js to the ProcessInfo servlet asynchronous.


tsurdilo:
+1

PR #450, opened 2016-09-09
[6.5.x] RHBPMS-4203 - Click on Process Documentation menu in IE11 and Chrome …

tsurdilo:
…almost entirely hide system buttons

toc plugin was causing problems inside designer frame. removing it


jlindop:
+1

PR #458, opened 2016-10-04
JBPM-5354 - Support case definition as new resource type in authoring

mswiderski:
@tsurdilo @jomarko could you please have a look?

### DIFF
--- a/jbpm-designer-client/src/main/java/org/jbpm/designer/client/handlers/NewCaseDefinitionHandler.java
+++ b/jbpm-designer-client/src/main/java/org/jbpm/designer/client/handlers/NewCaseDefinitionHandler.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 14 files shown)


mswiderski:
@tsurdilo this PR still missing the other property we talked about (caseRoles)

mswiderski:
@jomarko thanks, done

jomarko:
+1 from me

tsurdilo:
+1

jbpm-designer-client/src/main/java/org/jbpm/designer/client/handlers/NewCaseDefinitionHandler.java
```diff
@@ -0,0 +1,98 @@
...
+        designerAssetService.call( new RemoteCallback<Path>() {
+            @Override
+            public void callback( final Path path ) {
+                presenter.complete();
+
+                notifySuccess();
+                final PlaceRequest place = new PathPlaceRequest( path );
+                placeManager.goTo( place );
+            }
+        }, new DefaultErrorCallback() ).createCaseDefinition( pkg.getPackageMainResourcesPath(), buildFileName( baseFileName,
```
jhrcek: I know this has been merged already, but just for future PRs: using `new DefaultErrorCallback()` should no longer be necessary as it is used as default uncaught exception handler (see [here](https://github.com/droolsjbpm/kie-wb-common/pull/462/files#diff-a16fcaf1425005de11fd05cffed014c7R65))

mswiderski: @jhrcek do you mean I can simply skip the error callback? if so that's nice I didn't know about it :) thanks

mswiderski:
@manstis thanks for details. When it comes to not being communicated clearly, I think it might have been though I usually read that through but it does not mean I'll remember it if not practicing it every day. So I really appreciate this information here as that way I can improve next time as it was on top of my actual work with references which helps (at least me) to remember that (for a bit longer than after reading email :))

PR #462, opened 2016-10-05
JBPM-5354 - Support case definition as new resource type in authoring…

tsurdilo:
… -- adding case roles property


mswiderski:
looks very good to me +1

PR #463, opened 2016-10-06
JBPM-5352: root elements IDs are valid for BPMN2 modeler now

jomarko:


tsurdilo:
@jomarko : I'm not sure if generating "random" id with getRandomIdForRootElement is good here. 
We do have the org.jbpm.designer.util.Utils.toBPMNIdentifier util method already so running the msgID through that should be sufficient..or in the revisit method run 
signal.setId("_" + UUID.nameUUIDFromBytes(signal.getName().getBytes("UTF-8")));
through it to make sure it has the proper characters removed.  

also could we make sure (with test maybe) that this does not break reading of existing processes with messages created before this change?

jomarko:
@tsurdilo I updated the PR according to your suggestions.

However, tests for reading processes with messages are already present so I didn't add new one.

tsurdilo:
+1

PR #464, opened 2016-10-06
RHBPMS-4302: root elements IDs are valid for BPMN2 modeler now

jomarko:
Cherry-pick of #463


krisv:
Rather than trying to change the output of the designer, Bob is looking into relaxing the constraint on the signal id, as what designer is generating seems valid.

PR #466, opened 2016-10-07
[6.5.x] JBPM-5358 - Non escaped characters (&, <, >) in CDATA section in BPMN…

tsurdilo:
…2 result in invalid SVG file being generated


jlindop:
+1

PR #468, opened 2016-10-07
JBPM-5354 - Support case definition as new resource type in authoring…

mswiderski:
… - autostart property on task nodes

@tsurdilo @jomarko could you please have a look?


tsurdilo:
+1

PR #470, opened 2016-10-11
JBPM-5355 - Support case project - adding Milestone service task to d…

tsurdilo:
…efault wid with its default icon


jbpm-designer-client/src/main/resources/org/jbpm/designer/public/defaults/WorkDefinitions.wid.st
```diff
@@ -56,6 +56,15 @@ import org.drools.core.process.core.datatype.impl.type.ObjectDataType;
...
     "icon" : "defaultservicenodeicon.png"
+  ],
+
+  [
+    "name" : "Milestone",
+    "parameters" : [
+      "Condition" : new StringDataType()
+    ],
+    "displayName" : "Milestone",
+    "icon" : "defaultmilestoneicon.png"
```
mswiderski: just one comment - not sure milestone should be added to all none case projects as that's what we do when console creates new case project to provide special types there instead. But if we see a point in having it by default in all projects that's ok to.

mswiderski:
looks good to me - added one comment/question
+1

tsurdilo:
@mswiderski amended

mswiderski:
perfect, ready to be merged

PR #471, opened 2016-10-13
JBPM-5355 - Support case project - new case definition can be created…

tsurdilo:
… only in a case project


mswiderski:
+1

so now the case definition will only be available for case projects, excellent - thanks @tsurdilo

PR #473, opened 2016-10-18
JBPM-5355 - Support case project - add support for case files in proc…

tsurdilo:
…ess vars editor

@mswiderski - adding the "case file" option for process vars and case files section in the data io vars dropdown


mswiderski:
+1 works as expected, thanks @tsurdilo

PR #485, opened 2016-10-31
RHBPMS-4006 - Simulation on a process with a parallel gateway loop re…

tsurdilo:
…sults in StackOverflowError


jlindop:
Tested and looks good.
+1

PR #486, opened 2016-10-31
[6.5.x] RHBPMS-4393 - It is not possible to work with assets in Business Cent…

tsurdilo:
…ral if you generate form for task with unappropriated name


jlindop:
Tested and looks good
+1

PR #488, opened 2016-11-01
[6.5.x] RHBPMS-4365 - Scripts get reformatted after editing process causing c…

tsurdilo:
…ompile error


jlindop:
Tested and looks good.
+1

PR #489, opened 2016-11-01
RHBPMS-4365 - Scripts get reformatted after editing process causing c…

tsurdilo:
…ompile error


jlindop:
Tested and looks good.
+1

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonMarshaller.java
```diff
@@ -2202,7 +2202,7 @@ protected void marshallSubProcess(SubProcess subProcess, BPMNPlane plane, JsonGe
 
                 for(OnExitScriptType onExitScript : onExitExtensions) {
                     onExitStr += onExitScript.getScript();
-                    onExitStr += "|";
+                    onExitStr += "\n";
```
psiroky: This looks like something that could be refactored into a string constant (since it repeats multiple times).

PR #491, opened 2016-11-08
JBPM-5395 - rename ruleflow group query so it does not clash with stu…

tsurdilo:
…nner one plus fixes to ruleflow group editor


tsurdilo:
@jlindop this fixes the issue with name lookup for named queries between stunner and jbpm-designer for the ruleflow group query

jlindop:
Tested and looks good.
+1

PR #492, opened 2016-11-09
Deactive standalone showcase to fix build

csadilek:
@pefernan @psiroky of course, we need a permanent solution here. This is just to fix the build.


psiroky:
+1 as we need a working build. However, this need to be properly fixed.

csadilek:
I am closing this as we're working on a permanent solution for later today.

PR #493, opened 2016-11-09
RHBPMS-4483 - Process with any Subprocess are not valid

tsurdilo:


jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2UnmarshallingTest.java
```diff
@@ -1533,6 +1533,14 @@ public void testMITaskProperties() throws Exception {
     }
 
     @Test
+    public void testNoMIOnEmbeddedSubprocess() throws Exception {
+        Definitions definitions = loader.loadProcessFromJson("simpleEmbeddedSubprocess.json");
+        Process process = getRootProcess(definitions);
+        SubProcess subprocess = (SubProcess) getFlowElement(process.getFlowElements(), "abc");
+        assertNull(subprocess.getLoopCharacteristics().getId());
```
jomarko: I think the assert should looks like:
assertNull(subprocess.getLoopCharacteristics());

jomarko:
@tsurdilo Thanks for your PR. I am really sorry that this issue is caused by my previous PR. I found out this PR fine with only one suggestion above. Let me know what do you think.

jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2UnmarshallingTest.java
jomarko: @tsurdilo I agree that the current form of the assert is sufficient if we assume the current source code of unmarshaller. However, the unmarshaller can be changed in future. And if we use more restrictive assert (the one I have proposed) I think it can help to avoid similar issues (similar to this one I have caused by my last PR). What do you think?

tsurdilo: I think with the eclipse bpmn2 api the loop characteristics will currently not be null. Could we please +1 this to get it in as its urgent and test works and please add a commit that does what you would like to do after? WDYT?

tsurdilo:
@jomarko no problem at all! Let me know if doing your suggested change in a new pr would be good. I am on pto today and tomorrow so will not be able to change things until Monday. Thanks!

jomarko:
Sure, I am fine with separate PR. +1

PR #495, opened 2016-11-09
KIE forms jBPM integration

pefernan:
- Fixing build order
- Adding new modules


csadilek:
+1 (fixes build on master)

don't merge, only @psiroky can merge as build order has to change on jenkins.

PR #494, opened 2016-11-09
RHBPMS-4484 - Not valid process shows the same error message for diff…

tsurdilo:
…erent issues

PR #500, opened 2016-11-15
[6.5.x] RHBPMS-4484 - Not valid process shows the same error message for diff…

tsurdilo:
…erent issues (#494)


krisv:
+1

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/server/SimulationServlet.java
```diff
@@ -526,6 +526,11 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp)
                 resp.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
                 PrintWriter pw = resp.getWriter();
                 pw.write(e.getMessage());
+			} catch (Throwable t) {
```
psiroky: I know this was already merged, but still, should we really catch `Throwable` here? This will catch all the errors (like OOME) as well. That's not usually something we want.

PR #508, opened 2016-11-21
Removing Russian and Chinese (Traditional) properties files

paulovmr:
Since both locales were removed from zanata.xml file, they shouldn't come back when running a `mvn zanata:pull-module`.


paulovmr:
jenkins retest this

paulovmr:
jenkins retest this

PR #515, opened 2016-12-05
JBPM-5437: Escape quotes inside of entry/exit actions

jomarko:


tsurdilo:
@jomarko beat me to it :) thanks! 

+1

PR #516, opened 2016-12-09
JBPM-5459 - Add alternate ways to open the properties panel for nodes…

tsurdilo:
… and process properties


tsurdilo:
@jlindop 
ctrl-p opens/closes the property menu (right)
ctrl-m opens/closes the shape menu (left)

the shape menu icon opens/closes the property menu (right)

jlindop:
@tsurdilo The shape menu icon works for User Tasks, but for other objects (e.g. Send Task, End Events, etc), it looks as though the new PropseditButton is overlaid on top of the ShowSourceViewButton. Maybe the showPropseditButton function needs to adjust the button group number, as is done in e.g. showSourceViewButton.
Also I'm not sure how to make ctrl-p work - for me this opens the Print dialog in Chrome and Firefox.

tsurdilo:
@jlindop didnt have the problem with ctrl -p but will change that. also looking at the overlaying. thanks.

tsurdilo:
@jlindop amended (use ctrl+shift+p and ctrl+shift+m for the shortcuts now)

jlindop:
@tsurdilo  All working well now.
+1

PR #518, opened 2016-12-13
Change to use same Rhino version as Errai (old: 1.7R2 - new: 1.7R4)

paulovmr:


paulovmr:
Depends on: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/334

paulovmr:
Depends on: https://github.com/jboss-integration/jboss-integration-platform-bom/pull/343

paulovmr:
If someone could review this, the build on master is broken because the [another PR](https://github.com/jboss-integration/jboss-integration-platform-bom/pull/343) was already merged.

psiroky:
@paulovmr please note that we don't depend on IP BOM SNAPSHOTs, so the build should be fine. We should merge this only when we upgrade to the new IP BOM.

paulovmr:
@psiroky Oh yes, you are right. Probably failed for me because of some dependency cache from my tests. Thanks.

csadilek:
+1

PR #522, opened 2016-12-19
RHBPMS-4541 - Workflow with gateway is corrupting a Business Process

tsurdilo:
Should be "Workflow Patterns ...."
@jlindop you can test the "Synchronizing Merge" pattern as that one was for sure breaking things.


jlindop:
Tested and problem is fixed.
+1

PR #526, opened 2017-01-10
JBPM-5521 - Task name is not saved when you click in the swimlane

tsurdilo:
not sure why the diff on github is not showing correctly. small change to allow node name changes to happen when clicking on shapes (and still on canvas as well)


jlindop:
Tested and looks good. (I checked the code diff by comparing files locally).
+1

PR #528, opened 2017-01-13
JBPM-5552 - Remove the double border on subprocesses

tsurdilo:
@hasys mentioned subprocesses have a single border now - example image: http://imgur.com/a/Mg8Xl


jlindop:
+1

PR #529, opened 2017-01-16
JBPM-5478 - Data objects in swimlanes are removed on reopen

tsurdilo:


tsurdilo:
@jlindop data objects not allowed in container nodes -- https://streamable.com/c0g7p

jlindop:
+1

PR #532, opened 2017-01-19
UF-459: Preferences API Refactor

adrielparedes:
Preference API changed, so I needed to make some minor changes here.

Dependes on: https://github.com/uberfire/uberfire/pull/605


jlindop:
+1

PR #542, opened 2017-01-24
[UF-443] - Header and Footer should support Errai UI

ederign:
Depends on uberfire/uberfire#573

I'm getting an non related error when I try to run the WB:

https://gist.github.com/ederign/ccdbb182eaad66870af90222f7cedd47


csadilek:
+1

PR #544, opened 2017-02-01
JBPM-5635 - Drools language missing in an adhoc stage for completion …

tsurdilo:
…condition


jbpm-designer-client/src/main/resources/org/jbpm/designer/public/stencilsets/bpmn2.0jbpm/stencildata/bpmn2.0jbpm.orig
```diff
@@ -5578,12 +5578,17 @@
 							"value":"java"
 						},
 						{
-                                                        "id": "c2",
-                                                        "title":"javascript",
-                                                        "value":"javascript"
-                                                },
+							"id": "c1",
```
jlindop: New item for 'drools' language in stencilset has same id as 'java'

tsurdilo:
@jlindop amended. thanks.

jlindop:
@tsurdilo While testing this, I see that the script language property applies to the OnEntry and OnExit actions as well as the AdHocCompletionCondition. I wonder if it would be better to create a separate property for the CompletionConditionLanguage so that e.g the CompletionCondition could use 'Drools' while the OnEntry/OnExit scripts use 'Java'. What do you think?

tsurdilo:
@jlindop i agree with that. if ok lets create a separate jira for that so we can do this in both here and stunner. wdyt.

jlindop:
@tsurdilo Agreed. The PR looks good
+1

ibek:
thank you for fixing it so fast @tsurdilo  .... can we merge it now?

tsurdilo:
@csadilek could this be merged please? Same for https://github.com/droolsjbpm/jbpm-designer/pull/545. Thanks!

PR #547, opened 2017-02-10
[GUVNOR-2882] Rename Tomcat WAR, tomcat7 -> tomcat8

psiroky:
Needs to be merged together with https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/385


jbpm-designer-distribution-wars/src/main/assembly/assembly-standalone-tomcat-8_0.xml
```diff
@@ -3,7 +3,7 @@
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
           xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2 http://maven.apache.org/xsd/assembly-1.1.2.xsd">
 
-  <id>tomcat7.0</id>
+  <id>tomcat8.0</id>
```
manstis: As an aside; please see the comment [here](https://github.com/psiroky/kie-wb-distributions/blob/aef6c2d784797d0a8b2c7a282713eafaa748f96a/kie-drools-wb/kie-drools-wb-distribution-wars/src/main/community/assembly-kie-drools-wb-eap-7.xml#L22).. would it be an opportune time to fix this for designer?

psiroky: Good point. Be it my call, I would remove the distribution altogether. I doubt it even deploys anymore.

I'll at least fix the "dot" for now. Although I am not completely sure the comment applies anymore. In any case having just tomcat8 would be aligned with the rest, so will do that.

manstis:
+1 to merge.

PR #548, opened 2017-02-13
JBPM-5681 - DMN support for BusinessRuleTask

mswiderski:
depends on https://github.com/droolsjbpm/jbpm/pull/746

@tsurdilo do you mind having a quick look?

@etirelli this allows now to use DMN as rule language in designer

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
@@ -35,9 +35,11 @@ import org.jboss.drools.impl.DroolsFactoryImpl;
 import org.jbpm.bpmn2.xml.BPMNDISemanticModule;
 import org.jbpm.bpmn2.xml.BPMNSemanticModule;
 import org.jbpm.compiler.xml.XmlProcessReader;
+import org.jbpm.designer.bpmn2.BpmnMarshallerHelper;
 import org.jbpm.designer.web.profile.IDiagramProfile;
 import org.jbpm.process.core.validation.ProcessValidationError;
 import org.jbpm.ruleflow.core.validation.RuleFlowProcessValidator;
+import org.jbpm.workflow.core.node.RuleSetNode;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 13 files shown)


mswiderski:
the error is already described in droolsjbpm/jbpm#746 and should be fixed once this is merged https://github.com/droolsjbpm/kie-dmn/pull/38

etirelli:
@mswiderski it looks good, thank you for going the extra mile and doing this too! 

The only suggestion I have is that you are comparing strings like "DRL" and "DMN" in several places and then setting the value of attributes as the related enum like:

https://github.com/droolsjbpm/jbpm-designer/pull/548/files#diff-4e169fd35a32deb4eeb0e771fb9c92aeR1128

https://github.com/droolsjbpm/jbpm-designer/pull/548/files#diff-d32394cba930b1f8e7df123704c72537R4794

I usually prefer to centralize this kind of checks in the enum itself, because if in the future a string changes or new enum values are added, it is only necessary to change a single place, instead of searching across multiple projects to find where it is used. Example:

https://github.com/droolsjbpm/droolsjbpm-knowledge/blob/master/kie-api/src/main/java/org/kie/api/conf/EventProcessingOption.java#L74

https://github.com/droolsjbpm/droolsjbpm-knowledge/blob/master/kie-api/src/main/java/org/kie/api/conf/EventProcessingOption.java#L70

Other than that, +1 from me.

mswiderski:
jenkins retest this

mswiderski:
@etirelli I can move the DRL and DMN into constants like there are DMN_LANG and DRL_LANG in jBPM to improve this slightly. Though use of simpler values like DMN and DRL was to make it more user friendly instead of exposing URI which are less readable for average user. DMN and DRL is also defined in JS code

tsurdilo:
+1
thank @mswiderski very nice!

jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonMarshallerTest.java
```diff
@@ -446,6 +446,7 @@ public void testBusinessRuleTask() throws Exception {
         JSONObject properties = ruleTask.getJSONObject("properties");
         assertEquals("simpleGroup", properties.getString("ruleflowgroup"));
         assertEquals(true, properties.getBoolean("isasync"));
+        assertEquals("DRL", properties.getString("rulelanguage"));
```
jomarko: Please use appropriate constant from `BpmnMarshallerHelper`

jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonMarshallerTest.java
```diff
@@ -566,4 +567,12 @@ public void testAdHocSubprocessDroolsCompletionCondition() throws Exception {
         assertEquals("drools", scriptLanguage);
     }
 
+    @Test
+    public void testDMNBusinessRuleTask() throws Exception {
+        JSONObject process = loader.loadProcessFromXml("dmnBusinessRule.bpmn2");
+        JSONObject ruleTask = getChildByName(process, "test");
+        JSONObject properties = ruleTask.getJSONObject("properties");
+        assertEquals("DMN", properties.getString("rulelanguage"));
```
jomarko: Alse here, please use appropriate constant from `BpmnMarshallerHelper`

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
```diff
@@ -322,20 +324,52 @@ private void checkFlowNode(FlowNode flowNode, FlowElementsContainer container, P
...
+			}
+		} else if (ruleLanguage.equals(BpmnMarshallerHelper.RULE_LANG_DMN)) {
+			boolean missingNamespace = true;
+			boolean missingModel = true;
+			if (businessRuleTask.getIoSpecification() != null) {
+				List<DataInput> dataInputs = businessRuleTask.getIoSpecification().getDataInputs();
+
+				if (dataInputs != null) {
+					for (DataInput di : dataInputs) {
+						if (di.getName().equals("namespace")) {
```
jomarko: Whole `else` block for DMN is pretty defensive and checks for null values. Are you sure that di.getName() returns always non null value? I am just asking.

jomarko:
Hi @mswiderski, thanks for the nice PR, I put just some minor comments, could you please have a look?

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
```diff
@@ -322,20 +324,52 @@ private void checkFlowNode(FlowNode flowNode, FlowElementsContainer container, P
 	}
 
 	private void checkBusinessRuleTask(BusinessRuleTask businessRuleTask) {
```
jomarko: @mswiderski I had look also on the linked PR for jbpm. I see that there is done pretty same kind of validation like in this method. And there come idea to my mind. Isn't it possible to replace this method with something like [1]? Would be there many problems with error I18N and binding to specific node? What do you think @mswiderski @tsurdilo ?

[1]
https://github.com/mswiderski/jbpm-designer/blob/544bbd2f3b68a1826253f9468d215daa67d63a15/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java#L152

mswiderski: this is great idea @jomarko will update to use that instead, thanks

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
mswiderski: just checked the xsd and name of data input is optional so I'll update that .. if needed after using process validator

mswiderski:
@tsurdilo @jomarko updates done, thanks for review and please let me know if there are any other remarks

jomarko:
I remember now that I also wanted to reuse `RuleflowProcessValidator` in the past and now I remember why I decided not to do that. Now the errors related to `BusinessRuleTask` node thrown by `RuleflowProcessValidator` are bound to **process id**, however they should be bound to  **node id**. Because designer highlights nodes with errors according to ids bound to error messages. What more, the error messages related to DMN will appear only if `BPMN2SyntaxChecker` didn't find any other issues. Lets say there is script task with missing script and DMN business rule task with missing namespace. In this case `BPMN2SyntaxChecker` reveals script task errors and for this reason it doesn't call for RuleflowProcessValidator validation. In final, the designer doesn't show DMN related errors at all.

I thing there are two options:
# Bound error messages from `RuleflowProcessValidator` with proper id
# Throw error messages in the old way directly from `BPMN2SyntaxChecker` and in future do some larger refactoring that will enable to reuse `RuleflowProcessValidator` more reasonable

@mswiderski @tsurdilo what do you think?

@mswiderski I am sorry if I wasted your time and you will do it in the first way.

mswiderski:
fair enough @jomarko, will revert to use inline checks instead of RuleFlowProcessValidator

mswiderski:
jenkins retest this

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/validation/BPMN2SyntaxChecker.java
```diff
@@ -322,20 +324,52 @@ private void checkFlowNode(FlowNode flowNode, FlowElementsContainer container, P
...
+			}
+		} else if (ruleLanguage.equals(BpmnMarshallerHelper.RULE_LANG_DMN)) {
+			boolean missingNamespace = true;
+			boolean missingModel = true;
+			if (businessRuleTask.getIoSpecification() != null) {
+				List<DataInput> dataInputs = businessRuleTask.getIoSpecification().getDataInputs();
+
+				if (dataInputs != null) {
+					for (DataInput di : dataInputs) {
+						if ("namespace".equals(di.getName())) {
```
jomarko: Nice workaround 👍

jomarko:
+1 from me, thanks @mswiderski

PR #549, opened 2017-02-14
RHBPMS-4553 - jBPM process designer is not using the language chosen …

tsurdilo:
…in workbench settings

### DIFF
--- a/jbpm-designer-api/pom.xml
+++ b/jbpm-designer-api/pom.xml
@@ -43,6 +43,26 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>com.coderplus.maven.plugins</groupId>
+        <artifactId>copy-rename-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>copy-zanata-files</id>
+            <goals>
+              <goal>copy</goal>
+            </goals>
+            <configuration>
+              <fileSets>
+                <fileSet>
+                  <sourceFile>src/main/resources/org/jbpm/designer/resources/i18n/DesignerConstants.properties</sourceFile>
+                  <destinationFile>src/main/resources/org/jbpm/designer/resources/i18n/DesignerConstants_en.properties</destinationFile>
+                </fileSet>
+              </fileSets>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
 
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
@@ -62,6 +82,7 @@
             <configuration>
                 <includes>
                     <include>**/DesignerConstants.properties</include>
+                    <include>**/DesignerConstants_en.properties</include>
                     <include>**/DesignerConstants_de.properties</include>
                     <include>**/DesignerConstants_es.properties</include>
                     <include>**/DesignerConstants_fr.properties</include>
--- a/jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerWidgetView.java
+++ b/jbpm-designer-client/src/main/java/org/jbpm/designer/client/DesignerWidgetView.java
@@ -62,29 +62,17 @@ public class DesignerWidgetView
     public void setEditorID( final String editorID ) {
         this.editorID = editorID;
         inlineFrame.getElement().setId(editorID);
-        inlineFrame.getElement().setAttribute("name", editorID);
+        inlineFrame.getElement().setAttribute("name",
+                                              editorID);
     }
 
     @Override
     public void setEditorParamters( final Map<String, String> editorParameters ) {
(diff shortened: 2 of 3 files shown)


manstis:
Yay, my suspicion was right! 

Glad it was a simple fix.

tsurdilo:
@manstis i tested this a little more and have a question for you - when "english" is selected in the workbench as a language the locale is set to "default" but at that time the workbench somehow defaults to english and does not look at the JAVA_OPTS language settings. Do you know what this magic is ?

tsurdilo:
@manstis i found it - workbench sets english as the default language for "default". Now the question is what is the user expected behavior - to default to english or use the JAVA_OPTS settings users have defined ?

tsurdilo:
@manstis here i think "en" should be set and not "default" -- https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/configuration/popups/WorkbenchConfigurationPopup.java#L113

manstis:
@tsurdilo sounds reasonable if you've analysed the code. Maurico helped xiaibas(?) write these screens originally for jBPM Console. Much more than that I don't know!

tsurdilo:
@manstis i think we can tackle that in another jira (this one specifically requests for designer to behave like the workbench currently). I added a rename of DesignerConstants.properties as you did in GUVNOR-2722 and designer now defaults to en as workbench as well. 
I tested this on wildfly against master and 6.5.x branches and it works fine. With @jlindop being out now would you please review this PR? Thanks!

manstis:
@tsurdilo Have you checked the workbench **and** Designer with different locales (e.g Spanish and English)? I ask because the i18n properties file renamed for GUVNOR-2722 was **only** used for the login page and not elsewhere (I see your properties file looks to also be used by Oryx?). The i18n tag library used on the login page **always** appends the locale and hence the "default" (English) needed the locale appended. We do this for the ```kie-(drools)-wb``` distributions with some pom [magic](https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-drools-wb/kie-drools-wb-webapp/pom.xml#L1797).

tsurdilo:
@manstis yes i made sure switching workbench language propagates to designer as well. The main issue was setting for example: 
JAVA_OPTS="$JAVA_OPTS -Duser.language=de -Duser.country=DE"
and the workbench set to English.
In that case because workbench sets locale to "default" for English currently designer was using the java opts set locale and workbench was English but designer opened in German. Now both workbench and designer locales are aligned.

manstis:
@tsurdilo Great... now my next question ;-) 

"Locally" you have a ```DesignerConstants_en.properties``` file which is great for the taglib. However when a "push to Zanata" is performed the "base" file will also be ```DesignerConstants_en``` and subsequent "pulls from Zanata" will produce files such as ```DesignerConstants_en_de``` which the taglib won't like (won't match against a locale). This is why we copy and rename the files after fetching from Zanata for ```kie-(drools)-wb```. 

This operation is performed by @mbiarnes "Zanata" scripts on the CI server that handles Zanata changes. So... I suggest you liaise with @mbiarnes to emulate an update in Designer's i18n to check what affect the rename in this PR has on your i18n. I think you'll be surprised.

tsurdilo:
@manstis ok i will do that . thanks!

tsurdilo:
@mbiarnes could you please help me out with this? Basically we need to have a DesignerConstants_en.properties (currently there is just DesignerConstants.properties) - so what is the way to go about this? Do we replace DesignerConstants.properties with DesignerConstants_en.properties like I did in this PR, do we make a copy like @manstis showed in the previous comment or there is some other way to do this that I am not aware of? Thanks for any input!

tsurdilo:
@manstis updated to copy DesignerConstants to DesignerConstants_en in pom. This way we do not mess with zanata generated files. Tested and worst as expected. We should be good to merge now if you agree when build is stable again (jenkins build failing is not on designer module  - will ask to retest soon)

tsurdilo:
@manstis removed. thanks!

jbpm-designer-api/pom.xml
```diff
@@ -43,6 +43,26 @@
 
   <build>
     <plugins>
+      <plugin>
```
manstis: @mbiarnes Hi. You may need to add ```jbpm-designer``` to your i18n CI script to run this goal after fetching from zanata... @tsurdilo You may want to email Michael B about this; otherwise the copy may not run (there's a BASH script Michael has on the CI servers used for i18n synchronisation with Zanata; it runs the copy/rename plugin for kie-wb-distributions).

PR #551, opened 2017-02-15
JBPM-5460 - event subprocess - configurable abort propagation

mswiderski:
depends on https://github.com/droolsjbpm/jbpm/pull/749


jomarko:
Hi @mswiderski could you also add test into `Bpmn2UnmarshallingTest` that will verify that by default is the attribute true? I mean check for the `customAbortParent` also on some existing process where `customAbortParent` is not defined.

tsurdilo:
@jomarko 
the property definition element "default": "true" of customAbortParent  prop is going to set as default always in the json.

tsurdilo:
+1

mswiderski:
@jomarko it's actually not put into bpmn2 if it's the default as this property is considered optional so does not have to be given: https://github.com/droolsjbpm/jbpm-designer/pull/551/files#diff-d32394cba930b1f8e7df123704c72537R4917
it's consistent with how isasync is done so it reduces custom metadata if they are not needed - when defaults should be applied.

jomarko:
@mswiderski and @tsurdilo thanks for explanation. In that case I am also definitely for merging +1.

PR #560, opened 2017-03-25
JBPM-5843 - Updates for case definition modeling

tsurdilo:


tsurdilo:
@mswiderski @manstis could we merge please ?

PR #576, opened 2017-04-24
RHBPMS-4704 - Unable to open variable assignments screen in full scre…

tsurdilo:
…en mode


tsurdilo:
designer now puts the while page (including the workbench parts) into full screen mode instead of just its own frame.

PR #582, opened 2017-04-27
JBPM-5933 - Add option to not store / display bpsim info in bpmn2

tsurdilo:
To test:
start appserver with -Dorg.jbpm.designer.bpsimdisplay=false
when process is open:
1) properties window has no sim properties
2) Simulation tab is no present
3) simulation navbar functions are disabled (there but greyed out)
4) generated bpmn2 does not include the BPSim extension element (relationship) -- in view source as well as when saving it.


jbpm-designer-backend/src/test/java/org/jbpm/designer/web/server/SimulationServletTest.java
```diff
@@ -27,17 +27,25 @@
...
 import org.jbpm.designer.repository.UriUtils;
 import org.jbpm.designer.web.profile.impl.JbpmProfileImpl;
+import org.junit.Before;
 import org.junit.Test;
 
 import static org.junit.Assert.*;
 
 public class SimulationServletTest {
 
+    private JbpmProfileImpl profile = new JbpmProfileImpl();
```
jomarko: Could you please move this into the `setup()` method? It is more convenient to start each test with the 'clean' instance of `JbpmProfileImpl`

jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2UnmarshallingTest.java
```diff
@@ -1759,6 +1760,36 @@ public void testInputOutputSetsOfIoSpecification() throws Exception {
...
+        Definitions definitions = loader.loadProcessFromJson("userTask.json", "false", "true");
+        // should still be 0 even tho "true" is passed for bpsimDisplay
+        assertEquals(0, definitions.getRelationships().size());
+        System.clearProperty(EditorHandler.BPSIM_DISPLAY);
+    }
+
+    @Test
+    public void testEnableBpsimDisplayViaSystemProperty() throws Exception {
+        System.setProperty(EditorHandler.BPSIM_DISPLAY, "true");
+        Definitions definitions = loader.loadProcessFromJson("userTask.json", "false", "false");
```
jomarko: This brings me to idea, that the method `loadProcessFromJson` signature don't need to take also `bpsimdisplay` as an argument, because it can be easily set by system property (as you did). I think the less count of method attributes could help keep code readable, and setting the system property `EditorHandler.BPSIM_DISPLAY` is quite self explanatory enough I think. Let me know what do you think please.

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/profile/impl/EpnProfileImpl.java
```diff
@@ -1,361 +0,0 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-
-/**
- * The implementation of the epn profile for Process Designer.
- * @author Tihomir Surdilovic
- */
-@ApplicationScoped
-public class EpnProfileImpl implements IDiagramProfile {
```
jomarko: Could you please remove also `src/main/resources/org/jbpm/designer/public/profiles/epn.xml`?

jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2UnmarshallingTest.java
tsurdilo: removed

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/profile/impl/EpnProfileImpl.java
tsurdilo: removed now

jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/impl/Bpmn2UnmarshallingTest.java
tsurdilo: I think is ok as is because in tests we can "simulate" users having the default jbpm.xml (where bpsim display will be "true") and that via system property they can override that setting.
I have to add one more property like zorder and bpsimdisplay today for forms so will see if can simplify the method signatures not to have to take a bunch of true/false strings. Here I think its ok.

tsurdilo:
@jomarko thank you for the review! See replies to your comment regarding test methods. 
Also, I fully agree about the docs but I think designer docs need a lot more love than just adding this so I created JBPM-5947 for it.

jomarko:
Thanks @tsurdilo +1 from me.

PR #584, opened 2017-04-29
JBPM-5925 - Improve form generations from designer

tsurdilo:
- ftl form generation has been removed. 
- ftl inline form editor no longer used/exist
- designer no longer installs a ton of ".fw" form widgets into /globals dir 
- every form action (edit/generate) now asks user if they want form modeler or form modeler preview forms generated (.form, .frm)
- if you start appserver with -Dorg.jbpm.designer.formstype=frm or org.jbpm.designer.formstype=form the popup asking user for choice does not appear and always the option provided via this system prop will be used

@pefernan there seems to be an issue with "Edit Process Form" and "Edit Task Form" (for tasks/process where their form was not yet generated/does not exist). This goes into TaskFormsEditorServlet.getTaskFormFromRepository() correctly and if the form does not exist (which is fine) it gets the "empty" form value 

formValue = formBuilder.buildEmptyFormContent( taskName + TASKFORM_NAME_EXTENSION + "." + formType );

this form value looks like this for example:
{"id":"b28b6143-bcb3-4e6f-92ec-e492e2b60182","name":"PerformanceEvaluation222-taskform.frm","fields":[],"layoutTemplate":{"version":1,"layoutProperties":{},"rows":[]}}
and it is saved correctly.

The problem then is with the form modeler when form with this value is opened in workbench. It throws a JS error that pretty much stops any other assets from being able to be edited any more. 

@pefernan could you please look into this as it seems not related to designer.Thanks!


jhrcek:
@jsoltes @hasys this is important change in form generation, so be aware

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/server/TaskFormsEditorServlet.java
```diff
@@ -123,94 +126,109 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws S
...
-        String uniqueId = newFormAsset.getUniqueId();
-        if (Base64.isBase64(uniqueId)) {
-            byte[] decoded = Base64.decodeBase64(uniqueId);
-            try {
-                uniqueId =  new String(decoded, "UTF-8");
-            } catch (UnsupportedEncodingException e) {
-                e.printStackTrace();
-            }
-        }
+             String uniqueId = newFormAsset.getUniqueId();
```
jhrcek: This decoding logic is duplicated (I see it 3 times just in this PR). Could you please extract it to separate method and reuse?

jbpm-designer-client/src/main/resources/org/jbpm/designer/public/js/Plugins/inlinetaskformeditor.js
```diff
@@ -13,78 +13,207 @@ ORYX.Plugins.InlineTaskFormEditor = Clazz.extend({
...
+                });
+            }
+        } else {
+            Ext.Msg.show({
+                title : ORYX.I18N.inlineTaskFormEditor.formEditor,
+                msg : ORYX.I18N.inlineTaskFormEditor.selectForm,
+                buttons : {yes : ORYX.I18N.inlineTaskFormEditor.graphicalModeler, no : ORYX.I18N.inlineTaskFormEditor.graphicalModelerPreview, cancel : ORYX.I18N.Dictionary.cancel},
+                icon : Ext.MessageBox.QUESTION,
+                fn : function(btn) {
+                    if(btn == 'yes') {
```
hasys: Hi Tihomir,

if I correct the difference between 'Yes' and 'No' branches are "form" and "frm" literals. I guess we can handl it in other way than code duplication. Move duplicated logic to separate function or use ternary operator:

```javascript
    var editorType = (btn == 'Yes') ? "form" : "frm";
    if (action == "load") {
         this.showTaskFormEditor(editorType, options);
    } ...........
```

tsurdilo:
@jhrcek @hasys amended. thanks.

jhrcek:
@tsurdilo Thanks, +1

tsurdilo:
@jhrcek thanks  - can you under "Files Changed" approve via that "Review Changes" dropdown too pls?

tsurdilo:
@manstis i think this is good to be merged from designer point. @pefernan i think is still looking at the form modeler side regarding comment in this pr.

PR #591, opened 2017-05-08
[6.5.x] RHBPMS-4705 (6.4.z]  - Problem: Unable to open variable assignments screen in full screen mode in BPM Suite 6.4.x

tsurdilo:


tsurdilo:
@krisv 6.5.x port of jira - anyone that could approve?

krisv:
@tsurdilo I suggest you ask your QE contact to take a look and approve

tsurdilo:
@krisv i have asked QE in the JIRA. Thanks.

tsurdilo:
@hasys @jhrcek can we get this one looked at too when you have a chance ? Thanks!

hasys:
Hi @tsurdilo works good. Just two questions:
Maybe it won't be bad if Full Screen will be also a button to exit from Full Screen mode.
Also it is strange that in full screen Project Explorer going to "dark mode": 
image
I can reproduce it both in Google Chrome and in Firefox on Fedora 24.

tsurdilo:
@hasys we rely on the browsers "exit from full screen mode". Was like that before too. 
About the dark - thats a browser specific thing too. Not sure much I can change there. 

Thanks for reviewing!!!

tsurdilo:
@hasys I am looking into putting the entire html into full body mode instead of body element to see if that makes the fullscreen mode css not apply

tsurdilo:
@hasys the default background color of browsers full-screen is black. In order to change this the workbench should define in CSS rules to set the background color to what it should be 
The css that should be set are 
fullscreen
webkit-full-screen
moz-full-screen

Could we raise a workbench Jira for this please? I could force a change of the html/body element of the workbench and set a style in Designer but don't think that is a good idea as the workbench itself should define these styles for it.

PR #598, opened 2017-05-16
[7.0.x] JBPM-5989 - It is possible to create variables with the same name, bu…

tsurdilo:
…t different types


tsurdilo:
@hasys @jsoltes could you guys review? Thanks!!

PR #604, opened 2017-05-18
Removing wrong patternfly file

paulovmr:
This file was overwriting the correct one, that comes from Uberfire, on kie-wb-distributions.

@Rikkola @jhrcek This fixes the UI issue with the paginator on Library.

@cristianonicolai If you could take a look :)


jhrcek:

jhrcek:
@paulovmr thanks for the quick fix! I verified with kie-wb and @Rikkola now the pager really looks better (like on the pattern-fly catalogue)
screenshot from 2017-05-18 09-23-37

 Also many thanks to @psiroky who implemented support for full downstream build. This will make verification of PRs much easier

Rikkola:
image

Just showing here too what this fixes, in case we rebuild 7.0 final this should go in.

PR #608, opened 2017-05-22
JBPM-5984 - Workbench notify twice about broken Process

tsurdilo:
Correctly displays the notification now


hasys:

tsurdilo:
@manstis good point. i will create a jira to remove this class and just delegate. I believe this might have been added in early stages of integration with workbench where either a common service did not exist or i did not know about it. Either way I'll handle it in a new jira if thats ok. Thanks.

tsurdilo:
@manstis oh and i removed the iocmanager code from the class too per your request . thanks.

hasys:

tsurdilo:
@hasys that is weird because i am not getting the same --- https://streamable.com/y96wh

tsurdilo:
@hasys no problem at all - so are we good to merge (could you approve?) thanks!

PR #611, opened 2017-05-26
JBPM-5953: Add support for Service Task custom property values to Dat…

jlindop:
…a I/O editor

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
@@ -5990,15 +5990,8 @@ public class Bpmn2JsonUnmarshaller {
                     Assignment a = Bpmn2Factory.eINSTANCE.createAssignment();
                     FormalExpression fromExpression = Bpmn2Factory.eINSTANCE.createFormalExpression();
                     if (assignmentParts.length > 1) {
-                        // check if custom workitem property has been updated
-                        if (isCustomElement(properties.get("tasktype"),
-                                            preProcessingData) &&
-                                (properties.get(fromPart.toLowerCase()) != null && properties.get(fromPart.toLowerCase()).length() > 0)) {
-                            fromExpression.setBody(properties.get(fromPart.toLowerCase()));
-                        } else {
-                            String replacer = decodeAssociationValue(assignmentParts[1]);
-                            fromExpression.setBody(wrapInCDATABlock(replacer));
-                        }
+                        String replacer = decodeAssociationValue(assignmentParts[1]);
+                        fromExpression.setBody(wrapInCDATABlock(replacer));
                     } else {
                         // for custom workitem properties check individually for values
                         if (isCustomElement(properties.get("tasktype"),
--- a/jbpm-designer-client/src/main/java/org/jbpm/designer/client/shared/AssignmentData.java
+++ b/jbpm-designer-client/src/main/java/org/jbpm/designer/client/shared/AssignmentData.java
@@ -19,6 +19,7 @@ package org.jbpm.designer.client.shared;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -47,6 +48,7 @@ public class AssignmentData {
     private Map<String, String> mapDisplayNameToDataType = new HashMap<String, String>();
     private Map<String, String> mapDataTypeToDisplayName = new HashMap<String, String>();
     private Map<String, String> mapSimpleDataTypeToDisplayName = new HashMap<String, String>();
+    private Map<String, List<String>> mapCustomAssignmentProperties = new HashMap<String, List<String>>();
 
     private List<String> disallowedPropertyNames = new ArrayList<String>();
 
@@ -61,7 +63,8 @@ public class AssignmentData {
                           String sProcessVariables,
                           String sAssignments,
                           String sDataTypes,
-                          String sDisallowedPropertyNames) {
+                          String sDisallowedPropertyNames,
+                          String customassignmentproperties) {
         // setDataTypes before variables because these determine whether variable datatypes are custom or not
         setDataTypes(sDataTypes);
         setProcessVariables(sProcessVariables);
@@ -69,6 +72,7 @@ public class AssignmentData {
         setOutputVariables(sOutputVariables);
         setAssignments(sAssignments);
         setDisallowedPropertyNames(sDisallowedPropertyNames);
+        setCustomAssignmentProperties(customassignmentproperties);
     }
 
     public AssignmentData(String sInputVariables,
@@ -81,7 +85,8 @@ public class AssignmentData {
              sProcessVariables,
              sAssignments,
              null,
-             sDisallowedPropertyNames);
+             sDisallowedPropertyNames,
+             null);
     }
 
     /**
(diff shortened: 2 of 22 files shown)


tsurdilo:
@jlindop Hi, I ran your branch locally but seems nothing has changed (at least to me) from what was there before. Here is the update to the Email workitem i put in wid:

"parameterValues" : [
       "From" : "A,B,C",   
       "To" : "X,Y,Z",
       "Subject" : "one,two,three,four,five,six,seven,eight,nine,ten,eleven,twelve"
    ],

when i open the assignments editor and edit for example the "From" data input i don't see A B C options in the list. Do you think I am doing something wrong here?

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
```diff
@@ -5990,15 +5990,8 @@ protected void applyTaskProperties(Task task,
...
-                        // check if custom workitem property has been updated
-                        if (isCustomElement(properties.get("tasktype"),
-                                            preProcessingData) &&
-                                (properties.get(fromPart.toLowerCase()) != null && properties.get(fromPart.toLowerCase()).length() > 0)) {
-                            fromExpression.setBody(properties.get(fromPart.toLowerCase()));
-                        } else {
-                            String replacer = decodeAssociationValue(assignmentParts[1]);
-                            fromExpression.setBody(wrapInCDATABlock(replacer));
-                        }
+                        String replacer = decodeAssociationValue(assignmentParts[1]);
```
tsurdilo: This breaks the update of custom elements when "customEditor" : "true" is used in the wid. unless we completely disable this feature i think we should not break it :)

jbpm-designer-client/src/main/resources/org/jbpm/designer/public/stencilsets/bpmn2.0jbpm/stencildata/bpmn2.0jbpm.orig
```diff
@@ -3812,8 +3812,8 @@
 					"optional":true
 					 $if(workitemDefs.(k).customEditor)$
 					,
-					"hidden":true,
-					"visible":false
+					"hidden":false,
+					"visible":true
```
tsurdilo: why is this needed?

tsurdilo:
@jlindop looks good (see comments for some questions/ issues).
I just either don't know how to test it or there is something going on because i do not see the default values when a comma-separated string is used for parameterValues in the wid. Let me know if you can what i might be doing wrong here. thanks.

jbpm-designer-backend/src/main/java/org/jbpm/designer/bpmn2/impl/Bpmn2JsonUnmarshaller.java
jlindop: The change I've added to propertywindow.js (and function definitions in dataioeditor.js) ensure that it isn't broken. When the user sets a custom property, ORYX.DataIOEditorUtils.setAssignmentsPropertyForCustomAssignment is called which updates the 'assignments' property with the new custom property value setting. When the BP is serialized, the values for custom properties are taken from the 'assignments' property rather than from the custom properties themselves.

tsurdilo: @jlindop i tried used the per-property dropdowns when customeditor:true is set and it did not persist the values in my local testing but i can try again no problem.

jbpm-designer-client/src/main/resources/org/jbpm/designer/public/stencilsets/bpmn2.0jbpm/stencildata/bpmn2.0jbpm.orig
jlindop: This means that even when a service task has  customEditor set to true, the 'assignments' property is displayed, and the task has the button to bring up the DataIOEditor. Values for custom properties can be edited either in the custom property field or in the DataIOEditor, and the two are kept in sync. For example, if you update a custom property in the DataIOEditor, when the dialog is closed, the property field is updated with the new value.

jlindop:
@tsurdilo   The way I've implemented this, you still need to have "customEditor" : "true" set to enable parameterValues. I tried your example above for the email task and after adding "customEditor" : "true" it works fine.

jbpm-designer-client/src/test/java/org/jbpm/designer/client/popup/ActivityDataIOEditorViewImplTest.java
```diff
@@ -169,6 +175,25 @@ public void testSetPossibleOutputAssignmentsVariables() {
...
+        customAssignmentsProperties.put("To",
+                                        Arrays.asList(new String[]{"Tom"}));
+        customAssignmentsProperties.put("CC",
+                                        Arrays.asList(new String[]{"James", "John"}));
+
+        view.setCustomAssignmentsProperties(customAssignmentsProperties);
+        verify(inputAssignmentsWidget).setCustomAssignmentsProperties(mapCaptor.capture());
+        assertEquals(3,
+                     mapCaptor.getValue().keySet().size());
+        assertEquals(customAssignmentsProperties.get("CC"),
```
jomarko: I think there could be assertion also for "From" and "To"

jbpm-designer-client/src/test/java/org/jbpm/designer/client/popup/ActivityDataIOEditorTest.java
```diff
@@ -257,6 +263,23 @@ public void testProcessVariables() {
...
+        customAssignmentsProperties.put("TruckType",
+                                        Arrays.asList(new String[]{"Hiace", "Mazda"}));
+        customAssignmentsProperties.put("To",
+                                        Arrays.asList(new String[]{"Donna", "Diana", "Julia"}));
+
+        ioEditor.setCustomAssignmentsProperties(customAssignmentsProperties);
+        verify(ioEditorView).setCustomAssignmentsProperties(mapCaptor.capture());
+        assertEquals(2,
+                     mapCaptor.getValue().keySet().size());
+        assertEquals(customAssignmentsProperties.get("TruckType"),
```
jomarko: I think there could be assertion also for "To"

jbpm-designer-client/src/main/java/org/jbpm/designer/client/shared/AssignmentData.java
```diff
@@ -351,6 +356,68 @@ protected void setDisallowedPropertyNames(String disallowedPropertyNames) {
...
+                if (!propValues.isEmpty()) {
+                    Iterator<String> itPropValues = propValues.iterator();
+                    while (itPropValues.hasNext()) {
+                        String propValue = itPropValues.next();
+                        if (!propValue.isEmpty()) {
+                            sb.append(propValue).append(';');
+                        }
+                    }
+                }
+                sb.append(',');
```
jomarko: Just question here. This line appends the comma separator also after the last assignment. The method `setCustomAssignmentProperties` can handle this. Is this fine also for the javascript code?

jbpm-designer-client/src/test/java/org/jbpm/designer/client/popup/ActivityDataIOEditorViewImplTest.java
jlindop: Thanks - I'll add these.

jbpm-designer-client/src/test/java/org/jbpm/designer/client/popup/ActivityDataIOEditorTest.java
jlindop: I'll add this. Thanks.

jbpm-designer-client/src/main/java/org/jbpm/designer/client/shared/AssignmentData.java
jlindop: This implementation matches the javascript implementation here:
https://github.com/kiegroup/jbpm-designer/pull/611/files#diff-5b7904f8921bdfb9b4529d5bd6be3f36R218

The value of this string is only used by the Java code - it's passed from Javascript to Java when the DataIOEditor is invoked. On the return trip, it's only there for completeness of the serialization of the AssignmentData.

jomarko: thanks

tsurdilo:
@jlindop i tested again after watching your video and i think its pretty close thanks! Take a look at https://streamable.com/nvgdr if you have the time- if you set the standalone props without touching the Data io editor the data inputs double up (from 4 to 8 in this case). I'll debug and see what can be done.

jdlindop:
@tsurdilo  I've emailed you a fix for the bug in your video (:

tsurdilo:
i needed to make some minor fixes / changes but cannot amend this pr. will close this pr and open a new one that includes all this code and my changes.

PR #613, opened 2017-06-01
JBPM-5953 - Add support for predefined values to Data I/O editor

tsurdilo:
data i/o editor shows when customEditor is set to true in wid
data i/o editor contains and maintains the values of propertyValue entries for workitem data inputs (defined in the wid)

(testing propertyValue entries with comma-separated strings only - JBPM-5975 should bring the enum functionality solution in workbench - in tests this already works)

### DIFF
--- a/jbpm-designer-client/src/main/java/org/jbpm/designer/client/shared/AssignmentData.java
+++ b/jbpm-designer-client/src/main/java/org/jbpm/designer/client/shared/AssignmentData.java
@@ -19,6 +19,7 @@ package org.jbpm.designer.client.shared;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -47,6 +48,7 @@ public class AssignmentData {
     private Map<String, String> mapDisplayNameToDataType = new HashMap<String, String>();
     private Map<String, String> mapDataTypeToDisplayName = new HashMap<String, String>();
     private Map<String, String> mapSimpleDataTypeToDisplayName = new HashMap<String, String>();
+    private Map<String, List<String>> mapCustomAssignmentProperties = new HashMap<String, List<String>>();
 
     private List<String> disallowedPropertyNames = new ArrayList<String>();
 
@@ -61,7 +63,8 @@ public class AssignmentData {
                           String sProcessVariables,
                           String sAssignments,
                           String sDataTypes,
-                          String sDisallowedPropertyNames) {
+                          String sDisallowedPropertyNames,
+                          String customassignmentproperties) {
         // setDataTypes before variables because these determine whether variable datatypes are custom or not
         setDataTypes(sDataTypes);
         setProcessVariables(sProcessVariables);
@@ -69,6 +72,7 @@ public class AssignmentData {
         setOutputVariables(sOutputVariables);
         setAssignments(sAssignments);
         setDisallowedPropertyNames(sDisallowedPropertyNames);
+        setCustomAssignmentProperties(customassignmentproperties);
     }
 
     public AssignmentData(String sInputVariables,
@@ -81,7 +85,8 @@ public class AssignmentData {
              sProcessVariables,
              sAssignments,
              null,
-             sDisallowedPropertyNames);
+             sDisallowedPropertyNames,
+             null);
     }
 
     /**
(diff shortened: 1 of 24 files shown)


jomarko:
jenkins execute full downstream build

jbpm-designer-client/src/main/java/org/jbpm/designer/client/shared/AssignmentData.java
```diff
@@ -351,6 +356,68 @@ protected void setDisallowedPropertyNames(String disallowedPropertyNames) {
...
+                                }
+                            }
+                        }
+                        this.mapCustomAssignmentProperties.put(propName,
+                                                               listPropValues);
+                    }
+                }
+            }
+        }
+    }
```
tsurdilo: @hasys we could use streams but it would maybe just reduce couple lines. The issue is that this code is parsing the string values that have multiple separators and it just needs to do all those checks unfortunately imo.

jbpm-designer-client/src/main/java/org/jbpm/designer/client/shared/AssignmentData.java
```diff
@@ -351,6 +356,68 @@ protected void setDisallowedPropertyNames(String disallowedPropertyNames) {
...
+                        if (!propValue.isEmpty()) {
+                            sb.append(propValue).append(';');
+                        }
+                    }
+                }
+                sb.append(',');
+            }
+        }
+        return sb.toString();
+    }
```
tsurdilo: @hasys yeah same thing - i think could be reduced but parsing strings with couple separators is tricky

jbpm-designer-client/src/main/java/org/jbpm/designer/client/util/ListBoxValues.java
```diff
@@ -71,11 +69,26 @@ public ListBoxValues(final String customPrompt,
...
+        this.valueTester = copy.valueTester;
+        this.maxDisplayLength = copy.maxDisplayLength;
+        this.addValues(copy.acceptableValuesWithoutCustomValues);
+        if (copyCustomValues) {
+            for (String copyCustomValue : copy.customValues) {
+                this.addCustomValue(copyCustomValue,
+                                    null);
+            }
+        }
+    }
```
tsurdilo: @hasys yeah i think it is. in designer custom values are displayed as individual properties as well. when they change this copy should be made (old value really does not matter imo so its ok)

tsurdilo:
@jomarko fixed the issue with empty paramValue string (and added to test to make sure this works)
@jomarko regarding the cdata block -- from what I see there is no changes to the structure of the generated bpmn2. the from expression of the assignments contained it before as well so nothing should have changed but I agree adding more tests to make sure is good -- could you please create a jira to add those tests and I will do it asap.

tsurdilo:
@hasys I agree with you (regarding the customEditor may not be needed). This feature came from customer who did want to be able to turn it on / off depending on which of their users was using it so this would give them that ability. Maybe we should revisit in another jira? wdyt?

hasys:

jomarko:
@tsurdilo I filed two issues for test creation. Thanks.

hasys:

hasys:

tsurdilo:
@hasys after our chat today I have amended this pr with following changes:

1) customeditor : true is no longer needed in the wid for the Data I/O editor to recognize the paramValues options

2) Designer no longer displays the individual parameter entries for paramValues params and relies only on the data i/o editor. data i/o editor in this case is more powerful as it can map to the pre-selected values or user can also map to process vars etc if they want. 

@jomarko I will work on the 2 jiras you created this week - thanks!
@hasys I think this is ready to be reviewed again. Thanks.

tsurdilo:
jenkins please retest this

hasys:

tsurdilo:
@hasys are we ok now or do need to make any other changes?

hasys:
Hi @tsurdilo,
looks good. I am not able to check entire functionality due to JBPM-6069, but for now I have two questions/suggestions:
* Do we want to modify/add new predefined values in Data I/O editor?
* Do we want to remove support of customEditor? Probably, it can be used as hard coded predefined values, but Data I/O editor will be little bit more featured editor.

tsurdilo:
@hasys 
1) I think users can modify it in the wid. No need to also modify them inside the editor.
2) Yeah I think having both available is overkill. Before the data i/o editor was not shown when customEditor=true . Now that both are showing i think is worse becase dataio editor has now the capability to display the paramValues.

hasys:

hasys:

hasys:
Sorry, @tsurdilo,
I found an issue. If you click "Select previous version" you will see this:
image

tsurdilo:
@hasys looked into this and it's not related to this pr changes. we need a new jira for this error please. thanks!

tsurdilo:
the full downstream build is failing atm on the KIE-Maven Plugin module. Not related to this pr.

tsurdilo:

tsurdilo:

hasys:
Hi @tsurdilo,

I tried the latest master SNAPSHOT and yes, it looks like found issues is not related to this PR. I reported new issues here: https://issues.jboss.org/browse/JBPM-6078

This PR is ready to merge, thank you!

PR #614, opened 2017-06-07
JBPM-5975 Enumerations can't be used as predefined values

lanceleverich:
- Added direct dependency on org.eclipse.aether:aether-api
- Added direct dependency on org.kie:kie-ci
- Direct dependencies were added so that the workbench/Business Central can create the necessary ClassLoader, to be used when loading externally defined enumerations
- Added functionality to the evaluateWorkDefinitions method in JbpmPreprocessingUnit, to create a ClassLoader that can be used to load externally defined enumerations

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
@@ -24,6 +24,8 @@ import java.io.InputStream;
 import java.io.OutputStreamWriter;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
+import java.net.URL;
+import java.net.URLClassLoader;
 import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -47,11 +49,13 @@ import javax.servlet.http.HttpServletResponse;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.drools.compiler.kproject.ReleaseIdImpl;
 import org.drools.core.process.core.ParameterDefinition;
 import org.drools.core.process.core.datatype.DataType;
 import org.drools.core.process.core.datatype.impl.type.EnumDataType;
 import org.drools.core.process.core.impl.ParameterDefinitionImpl;
 import org.drools.core.util.MVELSafeHelper;
+import org.eclipse.aether.artifact.Artifact;
 import org.guvnor.common.services.project.model.Project;
 import org.guvnor.common.services.project.service.POMService;
 import org.guvnor.common.services.project.service.ProjectService;
@@ -75,6 +79,8 @@ import org.jbpm.process.workitem.WorkDefinitionImpl;
 import org.jbpm.process.workitem.WorkItemRepository;
 import org.json.JSONArray;
 import org.json.JSONObject;
+import org.kie.scanner.ArtifactResolver;
+import org.kie.scanner.MavenClassLoaderResolver;
 import org.kie.workbench.common.services.shared.project.KieProject;
 import org.kie.workbench.common.services.shared.project.KieProjectService;
 import org.slf4j.Logger;
@@ -526,6 +532,7 @@ public class JbpmPreprocessingUnit implements IDiagramPreprocessingUnit {
                                         String assetLocation,
                                         Repository repository) throws Exception {
         List<Map<String, Object>> workDefinitionsMaps;
+        List<URL> dependencyURLs = null;
 
         try {
             workDefinitionsMaps = (List<Map<String, Object>>) MVELSafeHelper.getEvaluator().eval(widAsset.getAssetContent(),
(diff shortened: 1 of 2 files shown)


lanceleverich:
Jenkins retest this please

tsurdilo:
@lanceleverich is it necessary to have to define mavenDependencies to the enum class? Typical scenario in workbench is that users just want to upload a jar file that includes their enums into their project. I guess question is if workbench does allow for auto-discovery of classes in a jar asset?

Also a test for this would be needed I think.

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
```diff
@@ -591,6 +598,25 @@ public void evaluateWorkDefinitions(Map<String, WorkDefinitionImpl> workDefiniti
                 } else {
                     workDefinition.setCustomEditor(null);
                 }
+                
+                // Get URLs so that we can create a URL-based ClassLoader, if necessary
```
tsurdilo: Could we please externalize this into a method that can be called if there is an enum type paramValue ?

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
```diff
@@ -618,9 +644,15 @@ public void evaluateWorkDefinitions(Map<String, WorkDefinitionImpl> workDefiniti
 
                 Map<String, Object> parameterValues = new HashMap<>();
                 if (workDefinitionMap.get("parameterValues") != null) {
+                    ClassLoader parentLoader = Thread.currentThread().getContextClassLoader();
```
tsurdilo: Does this really work when run in workbench? @manstis could you comment on this cause I am not sure.

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
```diff
@@ -618,9 +644,15 @@ public void evaluateWorkDefinitions(Map<String, WorkDefinitionImpl> workDefiniti
...
                     try {
                         Map<String, Object> parameterValuesMap = (Map<String, Object>) workDefinitionMap.get("parameterValues");
                         if (parameterValuesMap != null) {
+                            ClassLoader contextLoader = null;
```
tsurdilo: I assume most of paramValue entries would include comma-separated string of values, any way to only create this iff we are dealing with EnumDataType?

manstis:
FYI and as an aside IDK how Designer gets enum data BUT if it uses the DataModelOracle then the JAR containing the enum must be on the project's classpath. KieModuleMetaData uses the project classpath that includes Maven dependencies BUT I think falls back to the container classpath (so a JAR in /lib etc should also work - at least on WF/EAP).

manstis:
Uploading a JAR containing the enum to the Workbench (i.e. add new file and upload) was IIRC a 5.x feature. In 6.x you have to upload it to the Artifact Repository (ie Maven) and add as a project dependency IF using DMO.

tsurdilo:
@manstis thanks! ok so using mavenDepends and registering classes is the way to go then. I'll test this in workbench doing that.

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
tsurdilo: Yeah lines starting at 602 to 619

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
lanceleverich: The error was occurring because the EnumDataType could not find the enumeration/class that was used in the constructor of the EnumDataType. Even when the artifact was added as a dependency for the project, the Workbench's class loader did not include it; consequently it could not find the enumeration/class. The mavenDependencies would seem to be a logical place for adding a reference to the required artifact. Expanding on the parent class loader, and passing it to the EnumDataType.getValueNames() method, allows the enumeration/class to be found.

jbpm-designer-backend/src/main/java/org/jbpm/designer/web/preprocessing/impl/JbpmPreprocessingUnit.java
lanceleverich: I can refactor the code to only do the work associated with creating a new class loader when there is an EnumDataType.

lanceleverich:
Adding the artifact to the Artifact Repository (i.e. Maven) is not sufficient for the Workbench UI to pick it up. The WB's ModuleClassLoader (the class loader that is in-force when the system is evaluating/building work items) does not include items found in the project's dependencies, while it is in the process of evaluating/building work items.

manstis:
Perhaps not for Work Items (I did say IDK how Designer gets enums). But for *all* other editors we have a DataModelOracle that does get the enums from project dependencies. IDK what Stunner will use (if anything different to Designer) BUT the Workbench does have the means ALREADY to provide enum information to editors. Form Builder, Guided Rule, Guided Template, Guided Decision Table, Guided Decision Trees and Test Scenarios already use it.. I guess this PR could/should be considered a patch until (if!) Designer uses the DataModelOracle.

tsurdilo:
@lanceleverich could you please look into using DataModelOracle as @manstis mentioned and see if that is a viable option at this point? One thing I really thing is not good to have to tie paramValues in wid with mavenDependencies which is used for the workitem handler depends atm. wdyt?

tsurdilo:
Hi @lanceleverich this pr still needs a test and also conflicts resolved when you have the chance. Thanks.

lanceleverich:
Closing this pull request as the work that was being done here is now under a sub-task of the original JIRA-5975.

PR #622, opened 2017-06-20
Jbpm 6096

lanceleverich:
JBPM-6096 Create a solution based on obtaining a class loader
    * Created a unit test that checks for when mavenDependencies attribute is added to WID (allowing for a class loader that includes the external enum)
    * Created a unit test that demonstrates what occurs when external enum cannot be found, because it isn't in the current class loader's class path
    * Added a module to create the external enum so that it isn't automatically included in the class path of test


tsurdilo:

tsurdilo:
@mswiderski yes very good point! @lanceleverich the classloader search for depends is i think overkill after looking at this some more. All depends should be registered first and you can retrieve them via the guvnor project service POMService (look for example in ServiceRepoUtils on how the dependencies from this service are read and registered). This lookup should be a ton faster as well and would not introduce possible performance issues in future i think. thanks.

lanceleverich:
To address the concerns above...
1. The problem that is being addressed by the proposed fix is limited to the Workbench's BPM editor not being able to find the externally defined enumerations. The root cause of which I determined to be that when the EnumDataType.getValueNames method is called, it attempts to retrieve the values by using a Class.forName() call with the name of the externally defined enumeration class. However, that enumeration class does not exist in the class path, for the class loader that is operating at the time it is attempting to pre-process the WID. **NOTE: This is true even if you have added the jar as a dependency for the project.**
2. The mavenDependencies attribute of the WorkItemDefinition existed prior to me taking on this bug. I do not know what other use this attribute could have, if it was not to address this particular kind of issue.
3. Merging my code will not change the reporting that currently occurs, when the externally defined enumeration is not found.
4. Merging my code will only have any substantial impact if the WID includes a mavenDependencies attribute. If there are no mavenDependencies then it won't attempt to create a class loader.

tsurdilo:
mavenDepends prop were added to define the runtime dependencies of the workitem handler. During install of a workitem from service repo designer installs those for user in the workbench project. 
For design-time lookups we mentioned we would investigate further on what services the workbench can provide in order to make a decision how best to approach this. 

Keeping this pr open until we further investigate and have better understanding on best way to do it inside workbench.

tsurdilo:
@lanceleverich turns out we can use existing workbench services and project builder to do this -- https://github.com/kiegroup/jbpm-designer/pull/624 

i think using this integration is more suited for designer inside workbench than our initial thought of doing custom classloader and depends lookup from mavenDependencies. 

if you are ok with it we can close this pr (and associated subtask jira).

lanceleverich:
Closing this PR as @tsurdilo has submitted a solution that makes use of a pre-existing class loader.

PR #625, opened 2017-06-22
DROOLS-1608 - move processes related classes from drools to jbpm

tsurdilo:


tsurdilo:
@mswiderski @psiroky @mariofusco 

fix for designer after changes for DROOLS-1608

PR #628, opened 2017-06-29
JBPM-5602 - Process keeps the elements highlighted even though the va…

tsurdilo:
…lidation is off


tsurdilo:
@mswiderski this just makes sure that continuous validation is turned off and shape colors are set back before process is saved or closed.

PR #630, opened 2017-06-29
JBPM-5987 - It is possible to add stars as a defined type of a proces…

tsurdilo:
…s variable


tsurdilo:
@mswiderski just removing the silly "*******" divider in process var, globals editors and ruleflow group editor of businessRuleTask

PR #631, opened 2017-06-30
JBPM-5520 - Not able to delete elements using keyboard in a process a…

tsurdilo:
…fter change to other asset and back


tsurdilo:
@mswiderski just changing the keyboard binding for delete to meta-ctrl D instead of Delete key which the browser takes over after focus changes on the designer frame. On mac this is option-D.

mswiderski:
I guess we can't make test for it :)

PR #639, opened 2017-07-27
JBPM-6234 - Service task with nonstandard name is not displayed on th…

tsurdilo:
…e canvas


tsurdilo:
makes sure that only valid name property values from the wid get added as workitems. 

@hasys in addition to this i will add a new data type class similar to StringDataType in jbpm whihc allows users to specify a regex pattern as well. This will be done via a separate Jira.

PR #641, opened 2017-08-01
JBPM-6228 - Not possible to save project with any element in containe…

tsurdilo:
…r using Standalone mode


tsurdilo:
@hasys small fix to the save issue when container nodes are used

PR #651, opened 2017-08-29
RHBPMS-4836 - Classes from external POJO project not listed in variab…

tsurdilo:
…le definitions dropdown of the process modeler


tsurdilo:
@mswiderski do you mind looking at this when have the chance? Includes a test so no need to test in workbench (I have already done that)

PR #656, opened 2017-09-08
JBPM-6397 - Variables with same name but in different scopes causes I…

tsurdilo:
…llegalArgumentExceptions by indexer in designer

PR #657, opened 2017-09-11
[7.3.x] JBPM-6397 - Variables with same name but in different scopes causes I…

tsurdilo:
…llegalArgumentExceptions by indexer in designer (#656)


tsurdilo:
@mswiderski 7.3.x port

PR #688, opened 2017-10-24
JBPM-6509 - Support for sub cases both from within case and process

tsurdilo:


tsurdilo:
@mswiderski here you go :) thanks.

PR #698, opened 2017-11-14
DROOLS-2126: replacing old jackson 1.x with 2.x version

mareknovotny:
https://issues.jboss.org/browse/DROOLS-2126


mareknovotny:
there is another with the replacing jackson https://github.com/kiegroup/kie-wb-common/pull/1295

PR #701, opened 2017-11-23
JBPM-6642: removed jpdl resources

mareknovotny:
https://issues.jboss.org/browse/JBPM-6642

removal of jpdl resources which should be removed when JPDL feature was removed


tsurdilo:
@mareknovotny this was not removed before because if js plugin is not defined in jbpm profile it will never get loaded and was left there in case we needed this ever back. But +1 like this.

mareknovotny:
@tsurdilo ok, thanks for explanation, i was just looking at the list of issues I could do fairly easily myself ;) 
This is wanted to include in a product patch

PR #711, opened 2017-12-04
JBPM-6669 - Process name starting with number generates unexpected process id

tsurdilo:


tsurdilo:
@jomarko added negative test (to same test tho :) ). Thanks!

jomarko:
@tsurdilo hi, sorry for inconvenience, but the added negative test caused I double checked the specification, it looks like the NCName really can't start with digit, however `toIdentifierShouldChange` collection shows that this is still not accomplished. Could you please explain me a little bit? Or maybe we should ask some engine expert? @mswiderski.

tsurdilo:
@jomarko the method we use is to convert non-ncnname chars to their hexadecimal value which can then also include digits 0-9. So for example char '=' becmes '3d' etc. I think we should keep this approach but then remove all leading digits from the thenafter generated id. I will add that and update tests.

tsurdilo:
@jomarko updated code and tests. thanks!

PR #714, opened 2017-12-05
[7.5.x] JBPM-6669 - Process name starting with number generates unexpected process id

tsurdilo:


tsurdilo:
@jomarko updated per changes in the master pr after our discussions. thanks.

PR #716, opened 2017-12-08
[7.5.x] JBPM-6677 - Designer generates invalid task I/O when you open a proce…

tsurdilo:
…ss with task I/O containing underscore


tsurdilo:
@hasys project jira for https://issues.jboss.org/browse/RHBPMS-5077 - could you please review? Thanks

hasys:

mswiderski:
same here, is this needed for BRMS?

jsoltes:
@mswiderski @tsurdilo This fix should be merged also to 7.5.x, since human tasks are also available in RHDM. Another reason is that with RHDM GA we are also testing RHBA GA in parallel (this also is how this bug was discovered), so it makes sense also fixing the RHBA in 7.5.x

mswiderski:
@jsoltes what do you mean by this:

> since human tasks are also available in RHDM

I don't think that user tasks are available in RHDM ... unless I don't know what RHDM is ... User tasks require complete jBPM to be configured including data base etc.

jhrcek:
@mswiderski @tsurdilo  there's a [regression](https://issues.jboss.org/browse/RHBA-108) in designer reported for this  since 2 prod releases ago, so instead of reopening this, that one should be fixed instead.

jsoltes:
@mswiderski jhrcek's comment explains why I thought it is there (I just checked the library in the designer not knowing it is a bug). Still the second reason remains - we are testing RHBA in parallel with RHDM, so for the sake of faster feedback on the product quality it makes sense to include fixes for RHBA in 7.5.x, too.

PR #725, opened 2017-12-12
[7.5.x] JBPM-6694 - SVG process image in project with whitespaces in name

tsurdilo:


mswiderski:
is this needed for BRMS?

PR #737, opened 2018-01-23
JBPM-6768 - Designer says 'Succesfully generated process and tasks ta…

tsurdilo:
…skforms' even though they were not


tsurdilo:
@jsoltes @hasys do you guys mind taking a look? Thanks!

jsoltes:
jenkins execute full downstream build

hasys:
Hi @tsurdilo,

thank you for the PR. Just a question: any chance to show to the user the reason why it isn't possible to generate process and task form templates?

tsurdilo:
@hasys it's not clear from the exception thrown by the form service what the real reason is. We could show the full stack trace to the user but that to me is not very user friendly. They can see that in the logs. It would be nice for form service to do have like a validity check with errors returned then we could run that check and display issues to user. Designer just relies on the form service here so maybe a Jira to do something like that there would be nice to open. WDYT?

hasys:
Thank you @tsurdilo, looks reasonable. @jsoltes, do you know, do we have something similar to this issue? As requirements or bug report on forms side?

jsoltes:
@hasys I will report an enhancement for this

tsurdilo:
@jsoltes @hasys thanks guys. we can merge this i think - the failed test is not related to this

PR #742, opened 2018-02-06
JBPM-6805 - Designer produces XSD-invalid process XML if any node has…

tsurdilo:
… zero X or Y coordinate


tsurdilo:
@mswiderski could you please take a look? thanks

PR #740, opened 2018-02-01
JBPM-6794 - Invalid content was found starting with element caused by…

tsurdilo:
… 'documentation' ordering

PR #746, opened 2018-02-19
[7.5.x] JBPM-6794 - Invalid content was found starting with element caused by…

tsurdilo:
… 'documentation' ordering (#740)

### DIFF
--- a/jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/util/BaseElementFeaturesTest.java
+++ b/jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/util/BaseElementFeaturesTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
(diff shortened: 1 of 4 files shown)


tsurdilo:
depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/712

jbpm-designer-backend/src/test/java/org/jbpm/designer/bpmn2/util/BaseElementFeaturesTest.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
```
LuboTerifaj: Can you please change year to 2018?

tsurdilo:
@LuboTerifaj I think we should do this across the board and not just in this one file. I would leave this as is as the license header update is typically afaik a across-module global change that we do.

LuboTerifaj:
@tsurdilo I thought that when it is a new code, it could be right value. But I am ok with old value as well. Thanks!

PR #753, opened 2018-03-08
AF-1065 and AF-1061: New Process errors

paulovmr:
AF-1065: [Project Oriented] Unable to create new Process or Case Definition in the second project
AF-1061: [Project Oriented] Error dialog when creating Business process in new project after creating other project

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/repository/vfs/RepositoryDescriptorProvider.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/repository/vfs/RepositoryDescriptorProvider.java
@@ -67,10 +67,8 @@ public class RepositoryDescriptorProvider {
             branchName = "master";
         }
 
-        if (knownRepositories.containsKey(branchName + "@" + repositoryAlias)) {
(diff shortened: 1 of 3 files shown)


jbpm-designer-backend/src/main/java/org/jbpm/designer/repository/vfs/RepositoryDescriptorProvider.java
```diff
@@ -67,10 +67,8 @@ public RepositoryDescriptor getRepositoryDescriptor(Space space,
             branchName = "master";
         }
 
-        if (knownRepositories.containsKey(branchName + "@" + repositoryAlias)) {
-            return knownRepositories.get(branchName + "@" + repositoryAlias);
-        } else if (knownRepositories.size() == 1) {
```
ederign: this is ok to remove right?

paulovmr: @ederign This was causing the AF-1061 issue. I don't see a reason to keep it after the project oriented changes, since doesn't make sense to return any repository description just because it is the only one on this cache. The `else` clause after that should suffice.

PR #756, opened 2018-03-12
JBPM-6934 - Activation condition for case stage

mswiderski:
depends on https://github.com/kiegroup/jbpm/pull/1149


tsurdilo:
@mswiderski  I think pr is missing the adHocSubprocessActivationCondition.bpmn2 file.

mswiderski:
@tsurdilo you're right, just added them

PR #761, opened 2018-04-04
Removing redundant properties files

Rikkola:
This _en i18n file contains the same content as the default one.


manstis:
@Rikkola same commemt as on your other similar PR.

Rikkola:
Ok. Let's ask @tsurdilo he added this file and might remember the cause.

@tsurdilo There is some redundancy in the Constants files. The default and _en are covering the same scenarios. Do you remember if there was a reason to do this?

tsurdilo:
@Rikkola im removing it in this pr so maybe this one no longer needed: https://github.com/kiegroup/jbpm-designer/pull/763

Rikkola:
@tsurdilo Ok, thanks for the information. I think we should use `copy-rename-maven-plugin` also in kie-wb-distributions, if it is not already used. I'll take a look.

PR #762, opened 2018-04-05
JBPM-7093: Multiple Instances: Mapping automatically from MI data input/output to data I/O assignments (Human tasks)

mcivantos-tribalyte:
@krisv  @mswiderski This is the PR for MI data input/output assignments (Human tasks)


tsurdilo:
@mcivantos-tribalyte 
I see a bunch of System.out.println statements in the code that need to be removed. Also I would like to ask for more tests if possible. Also please test that if for any reason users already had defined a data input/output with the same name as what you are setting, that that will not be overwritten (or in case it has to be then make sure users cannot set data inoput/output with that name tobegin with in the data i/o editor). Thanks.

mcivantos-tribalyte:
@tsurdilo All the System.out.println statements are already deleted. Also, I've added a new test with the input/output data already defined. Do you need any other test?

tsurdilo:
@mcivantos-tribalyte ok thanks for doing that. i need to test this locally. will get to it this asap and update here with any findings. 

Just quick question I'm not sure why you need both
List<RootElement> rootElements = def.getRootElements();
and
List<RootElement> newRootElements = new ArrayList<RootElement>();?

mcivantos-tribalyte:
@tsurdilo I need to save the new rootElements items in a new array in order to updated def.getRootElements() at the end of the foreach loop, otherwise it returns an exception

PR #763, opened 2018-04-09
JBPM-7119 - Process Designer should save process before generating forms

tsurdilo:


tsurdilo:
@pefernan do you mind having a look please? If process is "dirty" users will now see this info message when trying to generate existing or create new forms:

screen shot 2018-04-09 at 12 30 52 pm

PR #775, opened 2018-04-30
RHPAM-674 - It is not possible to set constant value to Variable in I…

tsurdilo:
…E 11


tsurdilo:
@hasys could we get this reviewed please? its for blocker jira

PR #784, opened 2018-05-10
JBPM-7242 - Case Milestone do not support SLA duedate

tsurdilo:


mswiderski:
@tsurdilo has this been back ported to 7.7.x branch?

PR #786, opened 2018-05-11
JBPM-7265 - Add exporter attribute to default bpmn2 for new process a…

tsurdilo:
…nd new case process


tsurdilo:
@mswiderski this is needed after pr https://github.com/kiegroup/jbpm-designer/commit/bf49db93409914c904fed762db72ecaad97608df so that default process bpmn2 do not get "refreshed" after load.

PR #791, opened 2018-05-18
JBPM-7290 - Deselect all elements if json update is done when no expo…

tsurdilo:
…rter is defined


hasys:

PR #793, opened 2018-05-30
JBPM-7324 - Update process on import from bpmn2 if exporter is not de…

tsurdilo:
…fined


tsurdilo:
@mswiderski could you please have a look? Its JS changes only. Tested it in workbench works as expected. Thanks!

PR #802, opened 2018-07-03
JBPM-7408 - Correct id generation for default process and default cas…

tsurdilo:
…e project process

### DIFF
--- a/jbpm-designer-backend/src/main/java/org/jbpm/designer/server/service/DefaultDesignerAssetService.java
+++ b/jbpm-designer-backend/src/main/java/org/jbpm/designer/server/service/DefaultDesignerAssetService.java
@@ -347,18 +347,23 @@ public class DefaultDesignerAssetService
             location = location.replaceFirst("/",
                                              "");
         }
+
+        location = location.replaceAll("src/main/resources/", "");
+        location = location.replaceAll("/src/main/resources", "");
+
         location = location.replaceAll("/",
                                        ".");
-
-        if (location.length() > 0) {
-            String[] locationParts = location.split("\\.");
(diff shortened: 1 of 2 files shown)


tsurdilo:
@LuboTerifaj Hello again :) Mind taking a look at this as well? To test (this is master only):
1. Create new process in workbench
2. Look at the ID property (in property panel) of this process
before this change it used to be like "src.myprocess"
with this change it should include the package, for example "com.myproject.myprocess"
3. Make some changes to process and save it
4. Check the generated SVG file name. Before it was like "src.myprocess-svg.svg", now this should be fixed too and be for example "com.myproject.myprocess-svg.svg"

Thanks a lot.

tsurdilo:
@LuboTerifaj thanks for looking into this. Its for creating new processes to set the correct id. src.main is not a correct id for new created process. Some workbench changes in past changed how we retrieve paths so this is why i did this change to fix. What do you mean with "let it default" in your message. thanks

LuboTerifaj:
@tsurdilo
Thanks for clarification. 
With "let it default" I mean, that you can select package when creating new process. 
One option is "default", so if you select this one, id of process in (properties panel) is starting with "src.main".

tsurdilo:
@LuboTerifaj the <default> option cannot be selected - 
screen shot 2018-07-16 at 10 11 59 am

Am I missing something here?

tsurdilo:
@LuboTerifaj updated pr to add possibility of default package selection. The workbench will not allow selection of it any more (see previous comment). That might be a newer commit to master workbench branch so I added the fix anyways. Thanks.

LuboTerifaj:

PR #815, opened 2018-10-26
JBPM-7794 changing Case Definition label to legacy

tiagodolphine:
Changing Case Definition label to legacy, because now Stunner is going to be the default one.

@romartin 
@LuboTerifaj


tsurdilo:
@tiagodolphine can this be merged?

tiagodolphine:
Hi @tsurdilo yes, this is just a label change for Case Definition asset, now the default will be Stunner and jbpm-designer will be the "legacy".

PR #822, opened 2018-11-14
JBPM-7957 - Not able to set as reusable sub-process process from diff…

tsurdilo:
…erent project


tsurdilo:
@mswiderski do you mind taking a quick look please? i tested it locally and it works as expected. just JS changes here.

mswiderski:
@tsurdilo @hasys change itself looks ok but my question is about runtime ... since the subprocess comes from another project then that project must be added as dependency of this project to make this work on execution server. Plus configuration of kbase inclusion. So is this covered as well? If not then users would be able to select any process as called element but it will fail on runtime. With that said we should either query for processes of the project or projects added as dependency ... though it might not be that easy...

hasys:

krisv:
@tsurdilo @mswiderski I believe we already have that issue, as currently we have no efficient way of figuring out all the process ids included in this project or any of its (possibly transient) dependencies.  As a result, we used to just show all processes in the workbench.  If the user types in a process id, I guess it's currently his responsibility to make sure the runtime can find it.  I would love to see that workflow improved but that should then target the new designer imho.  Having it editable also allows other edge cases like using #{param} for dynamic resolution etc.

PR #826, opened 2018-12-04
workitem configuration fix

tsurdilo:


tsurdilo:
@mswiderski just a small fix for maven depends and removing useless default handler readonly param.

PR #849, opened 2019-05-14
JBPM-8479 - Sub-process variable lose it's type

tsurdilo:


tsurdilo:
@hasys  Hi, issue with this is how jbpm-designer (and idk if Stunner does the same but I assume it does) sets the  bpmn2:itemDefinition id element to the name of the variable. So if you have a process var and lets say 3 subprocesses all with same variable name, they will reuse the same itemdefinition. 
The way we got around this is to not allow a process variable (thats already set) to be used (its exact name) as process variables inside subprocesses. 
This check however only looked at process vars and not variables defined in other subprocesses. 
This small change here makes sure both process vars and already defined vars is subprocesses are also checked, so in your case you will not be able to have 2 subprocesses that have the same var names (the editor will show it as error and blank it out).

Changing the way jbpm-designer writes out the itemdefinition elements in bpmn2 would be a pretty significant change that I think if needed should only be done in Stunner.

tsurdilo:
Hi @hasys so what do you want to do with JBPM-8479 ? Close it ?

hasys:
Hi @tsurdilo,

yes, I guess it's better to close it to avoid issues with already created processes. Thank you!

@ibek, @krisv WDYT?

krisv:
Yes, unless there's a customer request to look into this, I would suggest to leave the old designer as is for now.

PR #867, opened 2020-02-06
latest Japanese translations from memsource

mbiarnes:


manstis:
@tsurdilo Hi.. so, for your information, we have a new process to deliver translations and they're being provided in UTF-8 format instead of ASCII encoded. I've confirmed this change breaks the login screen for Business Central however am less sure of the consequences to the old designer.. IDK if we even need to worry as isn't it removed from BC these days?

Can you advise?!?!

manstis:
@romartin FYI.. Do you know if we need to worry about legacy Designer i18n texts these days!?

mbiarnes:
@manstis like we agreed, I will close this PR because jbpm-designer will not be part of the BC  any-more

romartin:
Sorry for the late response @manstis @mbiarnes ! I was busy on 7.x and also sick for a few days...

Let you know that the (old / legacy) process designer is still included in the BC - The idea is not spending time on adding new features or refactoring it anymore, but for compatibility reasons, the legacy designer will still be included as part of the BC. It will be not possible to create new processes using the legacy designer, but it's inlucded becasue it's still possible to EDIT exiistng processes (being authoured with the legacy designer).

PR #870, opened 2020-02-18
Es fr translations

mbiarnes:


mbiarnes:
@manstis will close this and ask Terry for a new update only for jbpms-designer in ASCII format - we have to do still translations here since we were asked by Alessandro last time. If someone edits an existing process designed in an older version it seems that the jbpm-designer is still needed.

PR #881, opened 2020-05-13
AF-2410: default:// as main file system

adrielparedes:
## Intention

* Disable JGit specific features when JGit is not the default FS
* Add a switch to choose between Monitoring and Simplified Monitoring

## Simplified Monitoring 
Simplified monitoring is the K8S FS based solution. This can't use JGit specific parts so they are disabled (Like _Pages_)

Possible values:
* -Dorg.appformer.server.simplified.monitoring.enabled=true/false

## Related PRs:
* https://github.com/kiegroup/appformer/pull/882
* https://github.com/kiegroup/kie-wb-common/pull/3128
* https://github.com/kiegroup/drools-wb/pull/1363
* https://github.com/kiegroup/jbpm-designer/pull/881


tsurdilo:
LGTM :) If build is green go for it

